PR #2, opened 2017-02-07
Fix code formatting in README.rst

mythmon:
This fixes the code formatting of the following block of JS.


erikrose:
Looks like we pass in midflight. Fixed in 79725d689515343b133b43ec0ffcf617a50b58b7. Thanks, though!

PR #8, opened 2017-03-09
Add Travis support.

erikrose:
It's tricky to get Travis working on this, since we need both Node and Python.


krassowski:
What about installing `npm` with `sudo apt-get install npm` and then `npm install -g jsdoc`?

erikrose:
@krassowski Good idea! It worked!

Issue #9, opened 2017-03-09
support jsdoc config option (-c)


rhelmer:
All of the `jsdoc` command-line options can be set via config file.

`sphinx-js` should support a config value that can be specified in the sphinx `conf.py` which points to a `jsdoc` config file, such as `js_config_path`.

If `js_config_path` is not specified then the `-c` option should not be passed to the `jsdoc` command-line.

PR #12, opened 2017-03-10
Fix issue #9 - allow overriding jsdoc config file from sphinx conf.py…

rhelmer:
… via js_config_path.


erikrose:
Lgtm, except we should also add a mention of it in the README.

erikrose:
What do you think of #13? I went in to tweak the readme wording, and then I realized that the conf option was spelled kinda funny.

PR #15, opened 2017-03-25
[not ready for review] Tests for object initializer and private forward.

rhelmer:
Some code bases use "object initalizer" classes. A common approach
is to document the class in literal notation.
See
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

Also - some code bases use public classes which have public methods
that forward to a private class. The private class generally has the
jsdoc annotations. In this case, the documentation should use the
public class as the name, but take the documentation on the methods
in the private class.

### DIFF
--- a/tests/source/code.js
+++ b/tests/source/code.js
@@ -90,3 +90,111 @@ const NoParamnames = {};
 
 /** Thing to be shadowed in more_code.js */
 function shadow() {}
+
+/**
+ * Object initializer "class" definition in literal notation.
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer
+ * @class
+ */
+var ObjectLiteralClass = {
+    /**
+     * Foos the bars.
+     * @param {string} bar - the Bar to Foo.
+     * @returns {string} - Returns the Foo'd Bar.
+     */
+    foo(bar) {
+        return bar;
+    }
+}
+
+/**
+ * Private class renamed as public class, using the jsdoc lends tag.
+ *
+ * A common pattern in some codebases is to have a public class which
+ * forwards to a private class, where the jsdoc annotations live.
+ *
+ * @lends PublicClass
+ */
+class PrivateClass {
+    /**
+     * This is the method we want to appear on the public API.
+     * @param {string} foo - We want the foo.
+     */
+    public(foo) {
+        return foo;
+    }
+
+    /**
+     * This is a private method which should *not* appear on the public API.
+     * @param {string} foo - Gotta have that foo.
+     * @private
+     */
+    private(foo) {
+        return foo
+    }
+}
+
+/**
+ * This is the public API. All methods forward to PrivateClass.
+ */
+class PublicClass {
+    constructor() {
+        this._privateClass = new PrivateClass();
+    }
+
+    /**
+     * This should *not* appear in the documentation.
+     *
+     * @param {string} foo - You don't want this foo.
+     * @private
+     */
+    public(foo) {
+        return this._privateClass(foo);
+    }
+}
+
+/**
+ * Private class renamed as public class, using the jsdoc lends tag.
+ * Same as above but using object initializer written in literal object
+ * notation.
+ *
+ * A common pattern in some codebases is to have a public class which
+ * forwards to a private class, where the jsdoc annotations live.
+ *
+ * @lends PublicObjectLiteral
+ */
+var PrivateObjectLiteral = {
+    /**
+     * This is the method we want to appear on the public API.
+     * @param {string} foo - We want the foo.
+     */
+    public(foo) {
+        return foo;
+    },
+
+    /**
+     * This is a private method which should *not* appear on the public API.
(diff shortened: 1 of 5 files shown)


rhelmer:
I haven't narrowed down yet why the actual codebase I am testing against isn't working, in the meantime here are some reduced test cases to start with.

erikrose:
Just noticed this. I'm not sure why GH didn't notify me. Fff

tests/test.py
```diff
@@ -116,6 +116,25 @@ def test_getter_setter(self):
...
+            'autoclass_object_literal',
+            'class ObjectLiteralClass()\n\n   Object initializer "class" definition in literal notation.\n\n   ObjectLiteralClass.foo()\n\n      Foos the bars.\n\n      Arguments:\n         * **bar** (*string*) -- the Bar to Foo.\n\n      Returns:\n         **string** -- - Returns the Foo\'d Bar.\n')
+
+    def test_private_forwarding(self):
+        """Make sure public classes that forward to private classes can be documented."""
+        self._file_contents_eq(
+            'autoclass_private_forwarding',
+            'class PublicClass()\n\n   This is the public API. All methods forward to PrivateClass.\n\n   PublicClass.public(foo)\n\n      This is the method we want to appear on the public API.\n\n      Arguments:\n         * **foo** (*string*) -- We want the foo.\n')
+
+    def autoclass_object_literal_private_forwarding(self):
```
erikrose: Put a "test_" in front of this one so it runs, and it fails. :-(

rhelmer:
@erikrose - so these tests pass if I suppress https://github.com/erikrose/sphinx-js/blob/94307ed41ed5850430e6a1fc96bae3c8e642602b/sphinx_js/jsdoc.py#L50

The exception is:

```
Exception occurred:
  File "/Users/rhelmer/src/sphinx-js/sphinx_js/jsdoc.py", line 50, in run_jsdoc
    raise PathsTaken(conflicts)
sphinx_js.jsdoc.PathsTaken: Your JS code contains multiple documented objects at each of these paths:

  ./code.PublicClass#public
  ./code.PublicObjectLiteral.public
```

So what I'd expect in this case is for `PublicClass#public` to be ignored and not conflict here since it's marked `@private`. This appears to be what jsdoc does for its HTML output.

erikrose:
A couple of findings that reveal the depth of the depravity at issue:
* Remove the `@private` signifier from PublicClass.public and run jsdoc over it. Now you get both the shadowed and unshadowed versions in the HTML. There's no special handling of the intersection of privacy and lending; you're just seeing the usual effect of privacy hiding things. And, of course, we can't just uniformly ignore all private things, because often you do want to document them, as for methods meant for subclasses to call.
* In the `jsdoc -X` output, there is little indication of what shadows what. The only effect of `@lends` is to change the longname of the moved thing. The one hope I hold out is that we can infer what's shadowing what from the doclet's `meta.code.name` field. In the case of your example, it remains "PrivateClass#public" even when `@lends` changes the longname to "PublicClass#public". Perhaps we can figure out the general case. It's not immediately obvious what all the permutations are. For example, linkDensity has the name "linkDensity" but the longname ""linkDensity~length".

erikrose:
While trying to figure this out, I discovered an ironically undocumented tag you might use in the meanwhile: `@undocumented`. Apply it to the private things you want to ignore, and that dodges the ambiguity error.

erikrose:
Looks like meta.code.name should be expected to be a pathwise (not necessarily proper) suffix of longname. If it isn't, we're dealing something `@lend`-ed or `@alias`-ed (both of which share the same code path). Counterexamples are invited.

What's more, it turns out `@lends` has [the side effect of tagging the decorated object as `undocumented`](https://github.com/jsdoc3/jsdoc/blob/1c99ec39eb0fc61be627eec059d87348b8d7a93a/lib/jsdoc/tag/dictionary/definitions.js#L552). So one decent solution might be to notice which objects appear to be lent (because they're `undocumented` and don't conform to the suffix relationship above) and make sure they shadow conflicting entries in the suffix tree (as long as no other lent thing is already vying for the slot—then we really do have an ambiguity). How does that sound?

rhelmer:
> How does that sound?

sgtm!

erikrose:
Putting this aside for the moment, as we don't have a concrete situation that needs this: there are no methods in AddonManagerInternal which…
* Have a jsdoc comment and
* Are lent to AddonManager and
* Collide with a same-named method there which
* Also itself has a jsdoc comment.

In fact, that would be a weird situation to have: it's not unreasonable to have a comment on the proxy method which calls the lent method, but it would be weird to have a full-on jsdoc-formatted doclet there. That would imply that there are 2 distinct, noteworthy APIs, each worthy of being extracted into documentation, which I suggest raises the "proxy" method above the level of being a mere proxy.

So, for now, let's put this aside, since it adds nonobvious behavior to the code and is likely to prove brittle with future versions of jsdoc.

erikrose:
(Please reopen this if we find an applicable situation.)

PR #19, opened 2017-07-12
Use tempfile for subprocess stdout to handle large output volumes

jhkennedy:
I've been using sphinx-js to document the javascript part of my python package, and I ran into a issue where `sphinx_js/jsdoc.py`threw a `JSONDecodeError`. It turns out, the subprocess command to run jsdoc was producing a large output dictionary (149 KB; 5188 lines) that was only being partially returned leading to a syntax error (128K; 4474 lines). I've attached the two versions of the json file: 
[jsdoc-full.txt](https://github.com/erikrose/sphinx-js/files/1142968/jsdoc-full.txt) and [jsdoc-part.txt](https://github.com/erikrose/sphinx-js/files/1142971/jsdoc-part.txt) (as `*.txt` since github doesn't accept json files). 

When a subprocess command generates a lot of output to stdout (exceeds the OS pipe buffer), using `checkout_output` will end up returning an incomplete stdout (only what fits in the buffer).

Similar issues exist with Popen.communicate(), see:
* https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

The simplest solution is to use Popen instead of check_output and send stdout to a tempfile. Note: this will possibly also block for large stderr output, although (I believe) that is unexpected for a jsdoc call (a tempfile could also be used for stderr). See:
* https://stackoverflow.com/questions/3575554
* https://stackoverflow.com/questions/1180606

Note: messing with the `bufsize` option to `Popen()` didn't seem to help any.

### DIFF
--- a/sphinx_js/jsdoc.py
+++ b/sphinx_js/jsdoc.py
@@ -1,7 +1,8 @@
 from collections import defaultdict
-from json import loads
+from json import load
 from os.path import abspath, relpath, sep
-from subprocess import check_output
+from subprocess import Popen, PIPE
+from tempfile import TemporaryFile
 
 from .parsers import path_and_formal_params, PathVisitor
 from .suffix_tree import SuffixTree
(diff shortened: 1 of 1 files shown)


erikrose:
Thank you! We knew there would eventually be memory issues once jsdoc output got truly huge (we've been playing with using sphinx-js on Firefox itself), but I didn't expect them so quickly. I'll have a look.

erikrose:
Looks great. Thanks so much for the patch and for the thoughtful writeup that made it easy to review! If your documentation is public, I'd love to see it.

jhkennedy:
No, thank you for this package! It solved main main headache with our documentation which is huge. 

I'm using it in [LIVVkit](https://github.com/LIVVkit/LIVVkit), which is a verification and validation tool for ice sheet models. Our documentation is not quite complete yet, but you can see the docs for the last released version [here](https://livvkit.github.io/Docs/), and you should be able to build the latest set using the `generate_docs.sh` script in the [docs](https://github.com/LIVVkit/LIVVkit/tree/master/docs) directory. 

Note: there is a [LIVVkit.yml]() file you can use to setup an anaconda env for LIVVkit with most of the dependances (less jsdoc) using this command. You *should* just be able to:

```sh
git clone -b develop git@github.com:LIVVkit/LIVVkit.git
cd LIVVkit/docs
conda env create -f ../LIVVkit.yml
npm install -g jsdoc
./generate_docs.sh
```

And then the docs website will be at `_build/html/index.html`

erikrose:
Looks like a well-loved and well-documented project! :-D

PR #26, opened 2017-08-26
Add the ability to have multiple folders in js_source_path.

erikrose:
This is useful for slowly making the jsdoc cleanly parse in large projects like Firefox.

This still needs tests.


erikrose:
@rhelmer r?

PR #27, opened 2017-08-29
Handle encoding of temporary file

jhkennedy:
A temporary file is used to store the output of the `jsdoc` command in
case of a large output volume, but for some version of python and some
machines, the file encoding wasn't handled properly.

This commit explicitly opens a text-mode temporary file with `utf-8`
encoding which should eliminate this problem across pythons/platforms.

fixes erikrose/sphinx-js#20


erikrose:
Skillful and clever fix! Thank you! :-D

Issue #28, opened 2017-09-18
Conda error : "FileNotFoundError: [WinError 2] The system cannot find the file specified"


jeremysintes:
Dear all,

Thanks for this extension which looks amazing.

I think I may have an error related to the fact that I use anaconda. 
If you have any leads to solve it, it would be great!

Thanks in advance,
Jérémy

## Error message
`Running Sphinx v1.6.3
making output directory...
loading pickled environment... not yet created`

`Exception occurred:`
`  File "C:\Users\Jsintes\AppData\Local\conda\conda\envs\test_sphinx-js\lib\subprocess.py", line 957, in  startupinfo)`
`FileNotFoundError: [WinError 2] The system cannot find the file specified`

[sphinx-err-x_hvtdnq.txt](https://github.com/erikrose/sphinx-js/files/1311809/sphinx-err-x_hvtdnq.txt)

## Step to reproduce the error 
First I cloned the fathom project : 
`git clone https://github.com/mozilla/fathom`

Then created a conda env and activated it
`conda create -n sphinx-js`
`activate sphinx-js`

Then I installed the requirements
`npm install -g jsdoc`
`pip install sphinx`
`pip install sphinx-js`

And I tried to build the html from docs
`make html`

That's when the error occurred.

erikrose:
Is `jsdocs` (rather than `jsdoc`) above just a typo?

vladbytex:
Hello, 

I have a similar problem but on Ubuntu 17.04. When I try to run `make html` i get the following error:

`Sphinx version: 1.6.4
Python version: 2.7.13 (CPython)
Docutils version: 0.14 
Jinja2 version: 2.9.6
Last messages:

Loaded extensions:
Traceback (most recent call last):
  File "/home/vlad/.local/lib/python2.7/site-packages/sphinx/cmdline.py", line 305, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/home/vlad/.local/lib/python2.7/site-packages/sphinx/application.py", line 234, in __init__
    self._init_builder()
  File "/home/vlad/.local/lib/python2.7/site-packages/sphinx/application.py", line 312, in _init_builder
    self.emit('builder-inited')
  File "/home/vlad/.local/lib/python2.7/site-packages/sphinx/application.py", line 489, in emit
    return self.events.emit(event, self, *args)
  File "/home/vlad/.local/lib/python2.7/site-packages/sphinx/events.py", line 79, in emit
[...]

PR #29, opened 2017-09-19
Adding bug fix to enable running on windows

tim-foster:
Related to #28 . Right now sphinx-js does not run properly under windows.

I added the name variable from OS package. I then modify the Popen to add shell=True if we are running on windows. This fix allows me to run sphinx-js on a windows machine.

I have not been able to pass all unit tests. I am still working through the unit tests to see if any of them are failing because I am on windows.

I would be happy to provide more information or change anything if you feel fit.

Thank you,

Tim

### DIFF
--- a/sphinx_js/jsdoc.py
+++ b/sphinx_js/jsdoc.py
@@ -1,6 +1,7 @@
 from codecs import getwriter
 from collections import defaultdict
 from json import load
+from os import name as OSName
 from os.path import abspath, relpath, splitext
 from subprocess import PIPE, Popen
 from tempfile import TemporaryFile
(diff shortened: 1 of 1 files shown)


erikrose:
I'm not sure why Windows requires running under a shell, but I'm happy to accept the patch. (I'll probably tweak it to eliminate the repetition, but the basic approach seems fine.) Unfortunately, Travis doesn't do Windows builds, so I'll rely on you to tell me when you have all the tests passing, and then we'll merge this. Thanks for working on it! :-D

erikrose:
In 692ee1487d4f36d0d2ffe303fb70a829cb5c5c60, I removed the `stderr=PIPE` from the `Popen` call. If the trouble you were having was a hang, this should solve it. Can you please reopen this or comment if it doesn't work for you?

tim-foster:
I currently use a patched version for a workflow at work. I will test it with the latest master and check if that addresses the problem. Been meaning to address the unit tests but kept getting pushed down the road in priority.

tim-foster:
Yep. This addressed the issue. the stderr=PIPE has no impact on it. It was the jsdoc.cmd vs jsdoc.

erikrose:
Great! Thanks for getting back to me. I think the stderr bit helps only on large codebases where the stderr buffer gets full and then the OS and process deadlock, the OS waiting for the process to empty the buffer and the process waiting for the OS to close the pipe.

tim-foster:
Oh definitely. I was just curious if it directly impacted our situation. Great work. Thank you for this! It has saved so much time for documenting our code base and work flow.

PR #30, opened 2017-09-26
Add support to @typedef and @callback

krassowski:
Hi @erikrose,

Thank you for this sphinx extension. I have a tiny PR (only two lines changed) which makes a huge difference to real jsdoc documentation.

@[typedef](http://usejsdoc.org/tags-typedef.html) and @[callback](http://usejsdoc.org/tags-callback.html) are of kind 'typedef'. If we treat the `typedef` kind the same way as a `function` when parsing classes, we will get a nice documentation for **callbacks** (slightly changing just one line of `AutoClassRenderer.rst_for`!).

Here is an example - following declaration:

```js
    /**
     * Callback called after the plot is scrolled horizontally
     *
     * @callback position_callback
     * @memberOf NeedlePlot
     * @param {number} new_position - position (in aminoacids) of the first visible aminoacid in the plot
     * @param {boolean} stop_callback
     */
```
is currently shown as:
screenshot from 2017-09-26 21-06-50

and after the change it will become:
screenshot from 2017-09-26 21-07-11

The normal use of `typedef` can be accompanied with a lot of `properties` which could be really treated the same way as `params`. Please see the following example:

```js
    /**
     * @typedef {Object} Config
     * @memberOf NeedlePlot
     * @property {HTMLElement} element - an HTML wrapper in which the needleplot will be inserted
     * @property {Object} data - mutation and site data
     * @property {number} [width] - width in pixels
     * @property {number} [height] - height in pixels (or null if should be determined from ratio)
     * @property {NeedlePlot.Tooltip} [needle_tooltip]
     * @property {NeedlePlot.position_callback} [position_callback] - will be called on move event
     */
```

is now rendered as (no properties visible at all!):
screenshot from 2017-09-26 21-12-14

but after adding field type `properties` with the save formatter as `params` it becomes:

screenshot from 2017-09-26 21-20-56

I think that adding custom formatter for properties which would print out "Properties" instead of "Attributes" would be a better solution for the future but I feel a little bit uncomfortable editing larger chunks of the code now (it is difficult to understand some constructs with so little documentation). Also I would like to hear from you - if you are wiling to accept this patch or similar changes - before digging deeper in the code.

### DIFF
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -118,6 +118,7 @@ class JsRenderer(object):
 
         """
         FIELD_TYPES = OrderedDict([('params', _params_formatter),
+                                   ('properties', _params_formatter),
                                    ('exceptions', _exceptions_formatter),
                                    ('returns', _returns_formatter)])
         for field_name, callback in iteritems(FIELD_TYPES):
@@ -168,7 +169,7 @@ class AutoClassRenderer(JsRenderer):
 
         """
         def rst_for(doclet):
-            renderer = (AutoFunctionRenderer if doclet.get('kind') == 'function'
+            renderer = (AutoFunctionRenderer if doclet.get('kind') in ['function', 'typedef']
                         else AutoAttributeRenderer)
             # Pass a dummy arg list with no formal param list so
             # _formal_params() won't find an explicit param list in there and
--- a/tests/source/code.js
+++ b/tests/source/code.js
@@ -90,3 +90,15 @@ const NoParamnames = {};
 
 /** Thing to be shadowed in more_code.js */
 function shadow() {}
+
+/**
+ * @typedef {Object} TypeDefinition
+ * @property {Number} width - width in pixels
+ */
+
+
+/**
+ * Some global callback
+ * @callback requestCallback
+ * @param {number} responseCode
+ */
--- a/tests/source/docs/autofunction_callback.rst
+++ b/tests/source/docs/autofunction_callback.rst
@@ -1,0 +1,1 @@
+.. js:autofunction:: requestCallback
--- a/tests/source/docs/autofunction_typedef.rst
+++ b/tests/source/docs/autofunction_typedef.rst
@@ -1,0 +1,1 @@
+.. js:autofunction:: TypeDefinition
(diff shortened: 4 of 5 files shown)


erikrose:
Looks reasonable to me. How did you manage to implement `@callback` support without even mentioning "callback" in the code? Does jsdoc mix it down to a simpler construct?

krassowski:
> Does jsdoc mix it down to a simpler construct?

Well, yes and not, really both. [`@callback` is a synonym of the `@typedef`](https://github.com/jsdoc3/jsdoc/blob/1c99ec39eb0fc61be627eec059d87348b8d7a93a/lib/jsdoc/tag/dictionary/definitions.js#L829) but the `@typedef` is the "primary" tag. And [synonyms are "normalized" to](https://github.com/jsdoc3/jsdoc/blob/15b57e52f37d12dd059266c06702c830c0192191/lib/jsdoc/tag/dictionary.js#L104) (replaced with) the basal tags [for doclets generations](https://github.com/jsdoc3/jsdoc/blob/8b26b6d669741b1f3855e64eb836fc5ecddf320c/lib/jsdoc/tag.js#L162).

erikrose:
Thanks for clearing that up! I'm willing to merge this pretty much as-is as an experimental feature. All I ask is a test that covers `@typedef`. Basically, you add a new reStructuredText file to tests/source/docs, add a typedef to document in code.js, and add a test in test_build.py.

Callbacks are a good fit for function formatting, but it's unfortunate that the closest thing we have to work with for typedefs is `js:function`. Those `()` after the name and the "Arguments:" label are pretty misleading, as you point out. I've been taking advantage of Sphinx's existing js domain this whole time, but to make the formatting more appropriate to typedefs, you'd have to introduce a new kind of entity to the domain. Feel free to take a swing if you want, but you'll be blazing new territory. :-) Otherwise, I think it's better to land some support marked with caveats than no support.

krassowski:
I added two simple tests, both as sub-tests for autofunction.

> but to make the formatting more appropriate to typedefs, you'd have to introduce a new kind of entity to the domain. Feel free to take a swing if you want, but you'll be blazing new territory. :-) Otherwise, I think it's better to land some support marked with caveats than no support.

I agree! If I will need to document any of future projects with sphinx-js I may try to develop a nicer formatting for typedefs (and I will know where to start from).

erikrose:
Tests look great. I've merged it all in and released 2.2 so you don't have to work off a git checkout. Thanks again!

PR #33, opened 2017-10-17
Add "*" (and the rest) option to autoclass :members:.

mythmon:
This should allow me to control the ordering like I want in mozilla/pioneer-studies-addon-utils#4.

### DIFF
--- a/tests/source/code.js
+++ b/tests/source/code.js
@@ -48,6 +48,18 @@ class ContainingClass {
     }
 
     /**
+     * Another.
+     */
+    anotherMethod() {
+    }
+
+    /**
+     * More.
+     */
+    yetAnotherMethod() {
+    }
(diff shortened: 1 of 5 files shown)


tests/source/code.js
```diff
@@ -48,6 +48,18 @@ class ContainingClass {
     }
 
     /**
+     * Another.
+     */
+    anotherMethod() {
+    }
```
mythmon: I wanted a couple more methods in this class, to make my test for the feature convincing enough. I updated the tests that this affected as well.

erikrose: Yeah, I'm not proud of this shared test corpus. :-(

erikrose:
Looks great! Thanks especially for updating the readme as well!

PR #42, opened 2018-03-08
Render JSDoc @example tag.

### LINKED ISSUE #41: Support for @example?
Would you be opposed to supporting the [@example tag](http://usejsdoc.org/tags-example.html)? I took a quick look at it (since I'd like to keep as much inline in JS as possible, rather than in rST) and it seems fairly simple - simply read `examples` from the doclet in `_template_vars` and then extend the rST template.

lidavidm:
This does not render the optional caption. Let me know if you'd like to switch around the formatting; right now, all example sections are at the end.

Fixes #41.

Example:
screenshot 2018-03-08 at 12 36 46

### DIFF
--- a/sphinx_js/templates/function.rst
+++ b/sphinx_js/templates/function.rst
@@ -10,10 +10,12 @@
 
    {% if examples -%}
    :Examples:
-   {% for example in examples -%}
-     .. code-block:: js
 
-     {{ example|indent(5) }}
+   {% for example in examples -%}
+   .. code-block:: js
+
+      {{ example|indent(6) }}
+
    {% endfor %}
    {%- endif %}
 
--- a/sphinx_js/templates/function.rst
+++ b/sphinx_js/templates/function.rst
@@ -10,4 +10,11 @@
 
    {{ content|indent(3) }}
 
+   {% if examples -%}
+   :Examples:
+   {% for example in examples -%}
+     .. code-block:: js
 
+     {{ example|indent(5) }}
+   {% endfor %}
+   {%- endif %}
--- a/sphinx_js/templates/function.rst
+++ b/sphinx_js/templates/function.rst
@@ -8,8 +8,6 @@
      :{{ heads|join(' ') }}: {{ tail }}
    {% endfor %}
 
-   {{ content|indent(3) }}
-
    {% if examples -%}
    :Examples:
    {% for example in examples -%}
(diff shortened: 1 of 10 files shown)


lidavidm:
Ah I should come up with a test case, will do it soon.

erikrose:
Output looks great. What do you think of putting the examples before the ReST explanation? I think it's a little less surprising that way, since then the ReST will come at the bottom of the output, just as it comes at the bottom of the input.

sphinx_js/templates/function.rst
```diff
@@ -10,4 +10,11 @@
 
    {{ content|indent(3) }}
 
+   {% if examples -%}
+   :Examples:
+   {% for example in examples -%}
+     .. code-block:: js
 
+     {{ example|indent(5) }}
```
erikrose: Is 5 special? I see we used 3 in other places where we had to indent.

erikrose:
So yes, a test and reply about that 5 and maybe a mention in the readme, and we're ready to merge. Thanks for the good patch! :-D

sphinx_js/templates/function.rst
lidavidm: It's because I further indented the body of the `for` directive, so it needs an extra 2 spaces of indentation for everything to line up, with the `.. code-block::`. Do you think it's better to dedent it to stay consistent with the rest of the template?

lidavidm:
I debated whether the examples should go before or after the explanation. I do agree that it's probably better for the explanation to go last, so I'll change it.

lidavidm:
Changes:

- Moved rendered examples before the reST explanation (contrary to my commit message, whoops)
- Add tests for example tags
- Mention example tags in the README

lidavidm:
I also realized I had some warnings with build due to a missing blank line, and I've made the indentation consistent: it needs to be 3 + 3 more to indent the code under the `code-block`. I guess using the `:Example:` field list item isn't quite right though, would `**Example:**` be prefereable?

erikrose:
Changes look great. I agree, `**Example:**` is better. I took the liberty of doing it to save a back-and-forth. Merged! Thanks for the spiffy new capability! Keep 'em coming if you like; I love people who write their own tests. :-D

erikrose:
And [credited you in the readme](https://github.com/erikrose/sphinx-js/commit/fca5d2289fb66070445351114814d6d216e0237a).

erikrose:
If nothing else lands in the next few days, I'll roll a release.

lidavidm:
Thanks for the help, I'll definitely add more tags as I use them!

erikrose:
Excellent!

Issue #43, opened 2018-03-10
document.settings has no attribute 'tab_width'


jamrizzi:
I get the following error when using [sphynx-js](https://github.com/erikrose/sphinx-js).

```sh
Exception occurred:
  File "/home/jamrizzi/.local/lib/python3.6/site-packages/docutils/parsers/rst/__init__.py", line 171, in parse
    inputstring, tab_width=document.settings.tab_width,
AttributeError: 'Values' object has no attribute 'tab_width'
The full traceback has been saved in /tmp/sphinx-err-w9jvikut.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:20: recipe for target 'html' failed
```

PR #44, opened 2018-03-10
Fixed issue not attribute tab_width error

jamrizzi:
Fixes issue #43

### DIFF
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -23,6 +23,8 @@ class JsRenderer(object):
 
     """
     def __init__(self, directive, app, arguments=None, content=None, options=None):
+        if not hasattr(directive.state.document.settings, 'tab_width'):
+            directive.state.document.settings.tab_width = 4
         self._directive = directive
 
         # content, arguments, options, app: all need to be accessible to


erikrose:
Thanks for the report and the patch! Can you share with me a test case (even a manual one) that reproduces it?

jamrizzi:
```sh
git clone https://github.com/jamrizzi/sphinx-hello-world.git
cd sphinx-hello-world
make
```

erikrose:
Aha, it appears to be a new behavior that appeared sometime after Sphinx 1.6.5. If I build your project with that version, it builds fine. Continuing to chase...

jamrizzi:
Awesome, thanks

jamrizzi:
This issue has now bled into readthedocs builds. They must have updated the sphinx version about 1.6.5. It is urgent that this is fixed. Please accept my pull request. It fixes the bug.

sphinx_js/renderers.py
```diff
@@ -23,6 +23,8 @@ class JsRenderer(object):
 
     """
     def __init__(self, directive, app, arguments=None, content=None, options=None):
+        if not hasattr(directive.state.document.settings, 'tab_width'):
+            directive.state.document.settings.tab_width = 4
```
erikrose: Still trying to get to the bottom of why this fails. Anything against a default tab width of 8? I see that's the default in the docutils code.

erikrose:
I've got one more PR to review, and then I'll do a release. Thanks for contributing!

sphinx_js/renderers.py
jamrizzi: No prob, I don't really care what the default is. The default didn't noticeably effect anything.

jamrizzi: No prob, I don't really care what the default is. The default didn't noticeably effect anything.

PR #45, opened 2018-03-15
Fix line wrapping on windows

WimYedema:
Windows newline includes a LF(\r). This needs to be taken into
account when removing the wrapping.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@
 /.tox
 /dist
 /tests/source/docs/_build
-/build
-*/__pycache__/*
-/.vscode
 sphinx_js.egg-info/
+# Python 3
+*/__pycache__/*
+# Python 2.7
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -2,4 +2,7 @@
 /.tox
 /dist
 /tests/source/docs/_build
-/build
\ No newline at end of file
+/build
+*/__pycache__/*
+/.vscode
```
erikrose: Since it's editor-specific, this should probably be in your user's global .gitignore; then you won't have to repeat it for every project.

WimYedema: That's a good one. Didn't know I could do that...
The __pycache__ is created by python 3, python 2.7 generate .pyc files in the source. In the next patch I added that one too.

erikrose:
Thank you!

PR #46, opened 2018-03-15
Separate param type

WimYedema:
I split ':param type name: description' into ':param name: description' followed by ':type name: type'.
This allows the type to be more complex, for example including spaces, which is required for typescript types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 /dist
 /tests/source/docs/_build
 /build
+*/__pycache__/*
+/.vscode
+sphinx_js.egg-info/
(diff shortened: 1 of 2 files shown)


WimYedema:
Well... this patch builds on the previous.. probably better to abandon this one and create a pull request when the first one is merged?
Can you stack pull requests?

erikrose:
You can stack them, but it's kind of bad form if avoidable. However, stacking them is better than not submitting them at all. :-)

WimYedema:
Not to worry, I have a number of commits I want to push:) I thought to break them up into small commits so they are easily reviewed, but apparently it's also a bit more of a hassle...

PR #48, opened 2018-03-19
Fixes path on Windows

### LINKED ISSUE #47: Path resolution issue on Windows
Hello,

I have an issue for building a doc on Windows using sphin-js (it works on Linux and probably on Mac OS too).

Here is an example that will cause issue on Windows:

```rst
.. jsdoc:autoclass:: foo/bar/baz.Baz
```

Trying to build it on Windows:

```
$ make html
[...]
No JSDoc documentation was found for object "foo/bar/baz.Baz" or any path ending with that.
```
The problem is that Windows uses  `"\"` as path separator instead of `"/"`, so jsdoc reports paths like `"C:\...\foo\bar\baz.js"` in its JSON and sphinx-js does not handle this (it mixes both slashes and backslashes in the paths).

I made a quick fix (hack?) to allow my team to workaround this issue: https://github.com/flozz/sphinx-js/commit/c3471e3ef9af34ef7b91bf09c7928b4e14c63f6a but I do not know if I should PR it on this project or if we must find a better way to handle backslash on Windows...

:)

flozz:
Resolve #47

### DIFF
--- a/sphinx_js/jsdoc.py
+++ b/sphinx_js/jsdoc.py
@@ -2,7 +2,7 @@ from codecs import getwriter
 from collections import defaultdict
 from json import load
 import os
-from os.path import abspath, relpath, splitext
+from os.path import abspath, relpath, splitext, sep
 from subprocess import PIPE, Popen
 from tempfile import TemporaryFile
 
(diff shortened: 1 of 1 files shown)


erikrose:
I also cannot test this very well, but let's merge it and see if someone can try it out afterward. I don't see how it can break anything on UNIXy OSes.

PR #51, opened 2018-03-21
Improve params

### LINKED ISSUE #49: Missing parameters when using es6 object destructuration in parameters
Hello,

I have an issue to document a function that uses object destructuration in the parameters:

```javascript
/**
 * @param foo
 * @param {Object} params
 * @param params.bar
 * @param params.baz
 */
function example(foo, { bar, baz }) {
    // ...
}
```

Output with sphinx-js:

capture d ecran de 2018-03-20 10-44-38

Output with jsodc:

capture d ecran de 2018-03-20 10-45-41

When using destructuration, the JSON outputted by jsodc looks like this:

```javascript
[
    {
        // ...
        "meta": {
            // ...
            "code": {
                // ...
                "paramnames": [
                    "foo",
                    ""
                ]
            }
        },
        "params": [
            {
                "name": "foo"
            },
            {
                "type": {
                    "names": [
                        "Object"
                    ]
                },
                "name": "params"
            },
            {
                "name": "params.bar"
            },
            {
                "name": "params.baz"
            }
        ],
        "name": "example",
        "longname": "example",
        "kind": "function",
        "scope": "global"
    },
    // ...
]
```

As [sphinx-js relies on the `paramnames`](https://github.com/erikrose/sphinx-js/blob/master/sphinx_js/renderers.py#L109) to generate the function signature, it is not able to display the destructured parameters.

Do you think that we can use the `params` section of the doclet instead of `meta.code.paramnames`?

flozz:
Hello,

This PR allows sphinx-js to use documented params in function signature (when possible, if there is no documented param, it will use the ones from the js code).

I also added tests to cover the case.

Resolves #49


tests/test_build/test_build.py
```diff
@@ -48,7 +48,7 @@ def test_autofunction_callback(self):
         """Make sure @callback uses can be documented with autofunction."""
         self._file_contents_eq(
             'autofunction_callback',
-            u'requestCallback()\n\n   Some global callback\n\n   Arguments:\n      * **responseCode** (*number*) --\n')
+            u'requestCallback(responseCode)\n\n   Some global callback\n\n   Arguments:\n      * **responseCode** (*number*) --\n')
```
erikrose: Great side effect here. :-)

tests/test_build/test_build.py
```diff
@@ -60,6 +60,18 @@ def test_autofunction_example(self):
             '      // This is the example.\n'
             '      exampleTag();\n')
 
+    def test_autofunction_destructured_params(self):
+        """Make shure that all documented params appears in the function
+        definition."""
+        self._file_contents_eq(
+            'autofunction_destructured_params',
+            u'destructuredParams(p1, p2)\n\n'
```
erikrose: Remove all but first "u". (Unicode strings which collapse into non-unicode ones at the tokenizer level yield unicode strings.)

tests/test_build/source/code.js
```diff
@@ -141,3 +141,11 @@ class ExampleClass {}
  * console.log(ExampleAttribute);
  */
 const ExampleAttribute = null;
+
+/**
+ * @param {Number} p1
+ * @param {Object} p2
+ * @param {String} p2.foo
```
erikrose: This deserves an example in the readme. Is this the usual approach to documenting destructured formal params in JSDoc?

sphinx_js/renderers.py
```diff
@@ -109,8 +110,20 @@ def _formal_params(self, doclet):
         """Return the JS function or class params, looking first to any
         explicit params written into the directive and falling back to
         those in the JS code."""
-        return (self._explicit_formal_params or
-                ('(%s)' % ', '.join(doclet['meta']['code'].get('paramnames', []))))
+        if self._explicit_formal_params:
+            return self._explicit_formal_params
+
+        # Try to use documented params (@param) first
```
erikrose: Note to self: this adds params from the @param tag iff they're not subfields of of a collision with an existing one.

erikrose:
Looks good to me. I have just that one question, and then I want to add a little polish and merge it. Thanks!

tests/test_build/test_build.py
flozz: You are right, I removed useless unicode literals :)

sphinx_js/renderers.py
flozz: This part is maybe overcomplicated... Should I rewrote this using a `for` loop to make it easier to read?

tests/test_build/source/code.js
flozz: I added examples, but feel free to fix my english :)

sphinx_js/renderers.py
erikrose: No, I think it's fine and probably has advantages in guaranteeing a correct termination condition; I'll just add a comment.

tests/test_build/source/code.js
erikrose: Better than my French! :-)

erikrose:
Great fix! Thanks for working with me to get it merged.

PR #53, opened 2018-03-23
Display params default values

### LINKED ISSUE #50: Display default value in function signature
Hello,

When using sphinx to document Python code, it displays default parameters in the function signature:

[capture d ecran de 2018-03-20 10-59-28](https://wanadev.github.io/pyguetzli/pil_image.html#pyguetzli.pil_image.process_pil_image)

but sphinx-js does not:

```javascript
/**
 * @param foo
 * @param {String} bar
 */
function example2(foo, bar = "default value") {
    // ...
}
```

capture d ecran de 2018-03-20 11-00-27

As this information is available in the doclet, do you think we can display it (this issue probably depends on the resolution of #49)

PS: sorry to open as many issues :)

flozz:
Hello,

This PR displays the default value of parameters in the generated doc:
```javascript
/**
 * Lorem ipsum dolor...
 * 
 * @param {String} [name="foobar"]
 */
function example(name="foobar") {
   // ...
}
```

capture d ecran de 2018-03-23 16-09-44

Resolve #50

### DIFF
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -1,6 +1,5 @@
 from collections import OrderedDict
 from re import sub
-from functools import reduce
 
 from docutils.parsers.rst import Parser as RstParser
 from docutils.statemachine import StringList
@@ -117,10 +116,13 @@ class JsRenderer(object):
         # explicit formal param. Even look at params that are really
         # documenting subproperties of formal params. Also handles params
         # default values.
-        params = ["%s=%s" % (param, value) if value else param for param, value in reduce(
-            lambda params, param: params + [param] if not param[0] in [i[0] for i in params] else params,
-            [(param['name'].split(".")[0], param.get('defaultvalue')) for param in doclet.get('params', [])],
-            [])]
+        params = []
+        used_names = []
+
+        for name, default in [(param['name'].split(".")[0], param.get('defaultvalue')) for param in doclet.get('params', [])]:
+            if name not in used_names:
+                params.append("%s=%s" % (name, default) if default is not None else name)
+                used_names.append(name)
 
         # Use params from JS code if there are no documented params:
         if not params:
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -115,11 +115,12 @@ class JsRenderer(object):
 
         # Harvest params from the @param tag unless they collide with an
         # explicit formal param. Even look at params that are really
-        # documenting subproperties of formal params.
-        params = reduce(
-            lambda l, v: l + [v] if not v in l else l,
-            [param['name'].split('.')[0] for param in doclet.get('params', [])],
-            [])
+        # documenting subproperties of formal params. Also handles params
+        # default values.
+        params = ["%s=%s" % (param, value) if value else param for param, value in reduce(
+            lambda params, param: params + [param] if not param[0] in [i[0] for i in params] else params,
+            [(param['name'].split(".")[0], param.get('defaultvalue')) for param in doclet.get('params', [])],
+            [])]
 
         # Use params from JS code if there are no documented params:
         if not params:
--- a/tests/test_build/source/code.js
+++ b/tests/test_build/source/code.js
@@ -155,3 +155,15 @@ function destructuredParams(p1, {foo, bar}) {}
  * @param {string} [p2="foobar"]
  */
 function defaultValues(p1=42, p2="foobar") {}
+
+/**
+ * @param {number} [p1=42]
+ * @param {string} [p2="foobar"]
+ */
+function defaultValuesOnlyInDoc(p1, p2) {}
+
+/**
+ * @param {number} p1
+ * @param {string} p2
+ */
+function defaultValuesOnlyInJs(p1=42, p2="foobar") {}
--- a/tests/test_build/source/code.js
+++ b/tests/test_build/source/code.js
@@ -149,3 +149,9 @@ const ExampleAttribute = null;
  * @param {String} p2.bar
  */
 function destructuredParams(p1, {foo, bar}) {}
+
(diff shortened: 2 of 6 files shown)


erikrose:
I'm busy with ajn intense design sprint this week, but I'll be back to review this afterward.

sphinx_js/renderers.py
```diff
@@ -115,11 +115,12 @@ def _formal_params(self, doclet):
...
         # Harvest params from the @param tag unless they collide with an
         # explicit formal param. Even look at params that are really
-        # documenting subproperties of formal params.
-        params = reduce(
-            lambda l, v: l + [v] if not v in l else l,
-            [param['name'].split('.')[0] for param in doclet.get('params', [])],
-            [])
+        # documenting subproperties of formal params. Also handles params
+        # default values.
+        params = ["%s=%s" % (param, value) if value else param for param, value in reduce(
```
erikrose: Haha, okay, this needs some intermediate naming to make it readable. :-) Maybe break something out as a function, maybe just do part beforehand and stick it in a variable, but it's a lot to keep in your head as one big blob.

For that matter, this could keep us from recalculating `[i[0] for i in params]` every time around the reduce loop.

tests/test_build/test_build.py
```diff
@@ -72,6 +72,15 @@ def test_autofunction_destructured_params(self):
             '      * **p2.foo** (*String*) --\n\n'
             '      * **p2.bar** (*String*) --\n')
 
+    def test_autofunction_defaut_values(self):
+        "Make shure params default values appear in the function definition."""
```
erikrose: shure→sure

tests/test_build/source/code.js
```diff
@@ -149,3 +149,9 @@ const ExampleAttribute = null;
  * @param {String} p2.bar
  */
 function destructuredParams(p1, {foo, bar}) {}
+
+/**
+ * @param {number} [p1=42]
```
erikrose: If possible, it would be nice to test one of these params with the default specified only in the JSDoc comment and the other with it only in the function definition. That way, we will notice if we (or jsdoc) ever changes its behavior.

tests/test_build/test_build.py
flozz: Oops, I always do this error ^^'

sphinx_js/renderers.py
flozz: You are right, it was funny to write but it is unreadable :)

flozz:
I made the requested changes :)

tests/test_build/test_build.py
erikrose: English is weird. :-)

tests/test_build/test_build.py
```diff
@@ -72,6 +72,35 @@ def test_autofunction_destructured_params(self):
...
+            '   Arguments:\n'
+            '      * **p1** (*number*) --\n\n'
+            '      * **p2** (*string*) --\n')
+
+    def test_autofunction_defaut_values_only_in_js(self):
+        """Make sure params default values appear in the function definition when
+        the default values only appears in Javascript Code (ES2015)."""
+        self._file_contents_eq(
+            'autofunction_default_values_only_in_js',
+            u'defaultValuesOnlyInJs(p1=42, p2=foobar)\n\n'
```
erikrose: Hmm, missing quotes here. I'm seeing what can be done about that…

flozz: It seems to be how JSDoc works :(

it does not quote strings (the only way is to document the param including the quotes → `@param [foo="bar"]` as I did in the previous test...)

erikrose: Yes, I think the best course is to let it go without quotes. I was concerned that it would be ambiguous with default drawn from a var like…

```js
const hork = "snam!"
/**
 * @param {number} [p1=42]
 * @param {string} [p2]
 */
function defaultValues(p1, p2=hork) {}
```

…but JSDoc doesn't notice the default at all in this case. It also doesn't notice more dynamic constructs like…

```js
function defaultValues(p1, p2=woop()) {}
```

So we'll leave it unquoted for now. In fact, I can't see any way to disambiguate, say, `true` from `"true"` from the sphinx-js side. We'd have to fix it in jsdoc.

flozz: I came to the same conclusion... The best thing to do is to document explicitly the default values and to open a ticket on jsodc repository

erikrose:
Thanks for another good addition! :-D

tests/test_build/test_build.py
erikrose: I was wrong! If you feed this in…

```js
/**
 * @param p1
 * @param p2
 */
function defaultValues(p1=true, p2="true") {}
```

…it actually does the right thing:

```json
[
    {
        ...
        "params": [
            {
                "name": "p1",
                "defaultvalue": true
            },
            {
                "name": "p2",
                "defaultvalue": "true"
            }
        ],
        "name": "defaultValues",
        "longname": "defaultValues",
        "kind": "function",
        "scope": "global"
    },
    ...
]
```

Digging in more....

erikrose: So, yes, I think we can assume that any string value under `defaultvalue` is indeed meant to be a string literal. Boolean values are meant to be booleans. And ints and longs are meant to be numbers. Compound types like dicts may be possible as well. So what we really want is…

```python
params.append('%s=%s' % (name, _python_to_js(default)) if default is not None else name)
```

Where `_python_to_js` is something like…

```python
def _python_to_js(python_value):
    """Return JS code representing the given Python-equivalent data structure."""
```

I'm not going to get around to that for awhile, so feel free to give it a try if you like.

erikrose: Opened #56 to track this.

PR #57, opened 2018-04-04
Types of extracted default args are unambiguous in doclets after all. Get them right in extracted formal param lists. Fix #56.

### LINKED ISSUE #56: Get types right on default values extracted from JS code
See https://github.com/erikrose/sphinx-js/pull/53#discussion_r179174244 and the read of that thread.

erikrose:
JSDoc doesn't extract compound default literals like objects or Arrays yet, so we don't test those.

@flozz Would you mind reviewing this? Thank you!

### DIFF
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from json import dumps
 from re import sub
 
 from docutils.parsers.rst import Parser as RstParser
@@ -118,11 +119,12 @@ class JsRenderer(object):
         # default values.
         params = []
         used_names = []
+        MARKER = object()
 
-        for name, default in [(param['name'].split('.')[0], param.get('defaultvalue'))
+        for name, default in [(param['name'].split('.')[0], param.get('defaultvalue', MARKER))
                               for param in doclet.get('params', [])]:
             if name not in used_names:
-                params.append('%s=%s' % (name, default) if default is not None else name)
+                params.append('%s=%s' % (name, dumps(default)) if default is not MARKER else name)
                 used_names.append(name)
 
         # Use params from JS code if there are no documented params:
--- a/tests/test_build/source/code.js
+++ b/tests/test_build/source/code.js
@@ -151,8 +151,10 @@ const ExampleAttribute = null;
 function destructuredParams(p1, {foo, bar}) {}
 
 /**
- * @param {number} [p1=42]
- * @param {string} [p2]
- * @param {string} [p3="true"]
+ * @param [a=42]
+ * @param [b=a string]
+ * @param [c]
+ * @param [d]
+ * @param [e]
  */
-function defaultValues(p1, p2="foobar", p3="true") {}
+function defaultValues(a, b, c="true", d=true, e=null) {}
--- a/tests/test_build/test_build.py
+++ b/tests/test_build/test_build.py
@@ -74,24 +74,16 @@ class Tests(SphinxBuildTestCase):
 
     def test_autofunction_default_values(self):
         """Make sure params default values appear in the function definition,
-        whether the defaults are defined in JSDoc or JS.
-
-        Unfortunately, there is no way to disambiguate strings from symbolic
-        values used as default args in JS code: for instance, ``true`` vs.
-        ``"true"``. JSDoc's doclets render them both without quotes. However,
-        you can work around this by specifying your defaults in the JSDoc
-        comment rather than the JS code, as we do with p3. It would be lovely
-        if JSDoc changed its behavior someday so we could know to put quotes
-        around "foobar" as well.
-
-        """
+        whether the defaults are defined in JSDoc or JS."""
         self._file_contents_eq(
             'autofunction_default_values',
-            u'defaultValues(p1=42, p2=foobar, p3="true")\n\n'
+            'defaultValues(a=42, b="a string", c="true", d=true, e=null)\n\n'
             '   Arguments:\n'
-            '      * **p1** (*number*) --\n\n'
-            '      * **p2** (*string*) --\n\n'
-            '      * **p3** (*string*) --\n')
+            '      * **a** --\n\n'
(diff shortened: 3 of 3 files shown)


flozz:
Hi,

I made some tests, and there still be some inconsistencies between the values JSDoc finds from the js code (when using es6 syntax) and values manually documented in doclets. Also there is issue to document "complex" types like objects, arrays, functions,... But there is nothing we can do in sphinx-js to fix that, only JSDoc can fix this.

## Sphinx .rst

```
.. js:autoclass:: Foobar
   :members: simple_type_implicite,
             simple_type_explicite,
             simple_type_explicite2,
             array_object_function_implicite,
             array_object_function_explicite,
             constants_implicite,
             constants_explicite
```

## Example .js file

https://gist.github.com/flozz/0cb94dbe6e79d43ba31c5727684d24fe

## Tests with simple types

```javascript
/**
 * If we let jsdoc extracting the default values (es6 syntax), the
 * result is ok for simple types (number, string, boolean and null)
 *
 * @param {*} a
 * @param {number} [b]
 * @param {boolean} [c]
 * @param {string} [d]
 */
simple_type_implicite(a, b=42, c=false, d="false") {
}
```
capture d ecran de 2018-04-05 10-33-43

```javascript
/**
 * If we document explicitely values, it does not works as expected:
 * If something looks like a boolean or a number, it is "casted", so we
 * cannot have a ``"42"`` or a ``"false"`` string
 *
 * @param {*} a
 * @param {number} [b=42]
 * @param {boolean} [c=false]
 * @param {string} [d=false]
 */
simple_type_explicite(a, b=42, c=false, d="false") {
}
```
capture d ecran de 2018-04-05 10-33-53

```javascript
/**
 * An other exemple with explicite documentation
 *
 * @param [a=42]
 * @param [b="42"]
 * @param [c=false]
 * @param [d="false"]
 */
simple_type_explicite2(a, b, c, d) {
}
```

capture d ecran de 2018-04-05 10-34-03

## Tests with more complex types

```javascript
/**
 * More complex types are not documented automatically...
 *
 * @param [obj]
 * @param [arr]
 * @param [fn]
 */
array_object_function_implicite(obj={}, arr=[], fn=function(){}) {
}
```

capture d ecran de 2018-04-05 10-34-12

```javascript
/**
 * ... And are "casted" as string when explicitely documented (except for
 * objects that does not displayed at all, jsdoc seems to dislike ``{}``
 * in @param)
 *
 * @param [obj={}] Param JSDoc JSON: ``{"name": "obj="}``?
 * @param [arr=[]]
 * @param [fn=function(){}]
 */
array_object_function_explicite(obj={}, arr=[], fn=function(){}) {
}
```

capture d ecran de 2018-04-05 10-34-17

## Tests with constants

```javascript
/**
 * Variable or constants as default value are not automatically
 * documented...
 *
 * @param  [a]
 */
constants_implicite(a=SOME_CONST) {
}
```

capture d ecran de 2018-04-05 10-34-31

```javascript
/**
 * ... and are seen as string if documented explicitely
 *
 * @param [a=SOME_CONST]
 */
constants_explicite(a=SOME_CONST) {
}
```

capture d ecran de 2018-04-05 10-34-38

flozz:
Sorry for the looooong message...^^'

**I do not know what is the best thing to do on sphinx-js side:**

* Trusting the types given by jsodc?
* Just write the default value as is?
* Trying to be smarter (looking at the param type when available)?

**On JSdoc-side:**

* Maybe we can also open an issue... but they will not be able to fix this behavior without breaking things...

Anyway, thank you for asking me to review that PR :)

erikrose:
To summarize (and get it straight in my head)…

Using the doclet to document defaults lets you use [], {}, function(), and other complex values. However…

* Interfacing with that the simple way, by merely inserting the doclet text verbatim into our generated params, conflicts with [the JSDoc docs](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values), which never use quotes for strings. We would have to use some heuristic, which destroys the unambiguity we seek.
* Using the declared type information to guide our interpretation of the default value would help for any value that can be read as JSON (false, 42, [], {}), but dynamic values like a function call that returns an int) would fail. I suppose, in case of failure, we could fall back to including the text verbatim. (JSDoc would also like to fix this on its side, but [it doesn't look like it will happen soon](https://github.com/jsdoc3/jsdoc/issues/902).)

Documenting defaults solely in the JS code is the other option. But this limits you to the types JSDoc can extract: true, 42, "string", and null.

And unfortunately, sphinx-js doesn't receive both the doclet and the JS default: we get only one. So the best course I see is to use the explicit type declaration, if any, to interpret. If we have a declaration and it is "string", put quotes around the literal text and stitch it into our template. Otherwise, omit the quotes. But this, too, has a problematic corner case: if the type declaration is a disjunction like `{Array|string}` and the default is `[]` (which comes through to sphinx-js as "[]", bizarrely), we won't know what to do. I suppose we could just take the first one and document our heuristic. I wonder if all that will actually work. Thoughts?

I also looked at the possibility of using workaround: for simple_type_explicite and array_object_function_explicite, I'd recommend leaving out the doclet bits that confuse JSDoc. I'm not sure where the `""` is coming from in simple_type_explicite2; that warrants investigation. For array_object_function_implicite, constants_implicite, and constants_explicite, I think all we can do for now is to document the default in the description part of the doclet, as in "Default: []".

erikrose:
Hmm, I just don't like making people deviate from stock JSDoc syntax; I want sphinx-js to be able to be dropped into existing projects. Let's let it bake a day or two.

erikrose:
How does this strike you, @flozz? It still needs documentation in the readme, but this way, you can express anything, ES2016-style defaults are always interpreted as correctly as JSDoc allows, and we don't deviate from JSDoc's documented spellings.

flozz:
For me, it is the right compromise. I tested on my project and it worked just fine :D

erikrose:
And I published [2.5 on PyPI](https://pypi.org/project/sphinx-js/). Cheers!

PR #64, opened 2018-05-20
Variadic parameters & @deprecated

lidavidm:
Adds support for `@deprecated` and variadic parameters. Once these look good, I can add the test cases.

I rendered variadics as a Python star-arg, but on second thought, it'd probably make more sense to change it to an ellipsis.

image

Corresponding JSDoc:

```js
/**
 * Given projection IDs, returns a function that always returns those
 * IDs.
 *
 * Useful as the ``childrenFunc`` argument to a view when your
 * children are static.
 *
 * @param {...*} projections - The view IDs. (This is a variadic
 * function.)
 *
 * @alias gfx.constant
 */
```

I've rendered `@deprecated` as a `note` instead of `deprecated`, since the latter expects a version and possibly descriptive text, while JSDoc only gives you a chance to put descriptive text. I'm open to other suggestions on how to render this.

image

Corresponding JSDoc:

```js
    /**
     * @deprecated Use :func:`allocate` instead.
     */
```

### DIFF
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -222,6 +222,7 @@ class AutoFunctionRenderer(JsRenderer):
             fields=self._fields(doclet),
             description=doclet.get('description', ''),
             examples=doclet.get('examples', ''),
+            deprecated=doclet.get('deprecated', ''),
             content='\n'.join(self._content))
 
 
@@ -234,6 +235,7 @@ class AutoClassRenderer(JsRenderer):
             params=self._formal_params(doclet),
             fields=self._fields(doclet),
             examples=doclet.get('examples', ''),
+            deprecated=doclet.get('deprecated', ''),
             class_comment=doclet.get('classdesc', ''),
             constructor_comment=doclet.get('description', ''),
             content='\n'.join(self._content),
@@ -314,6 +316,7 @@ class AutoAttributeRenderer(JsRenderer):
         return dict(
             name=name,
             description=doclet.get('description', ''),
+            deprecated=doclet.get('deprecated', ''),
             examples=doclet.get('examples', ''),
             content='\n'.join(self._content))
 
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -223,7 +223,7 @@ class AutoFunctionRenderer(JsRenderer):
             description=doclet.get('description', ''),
             examples=doclet.get('examples', ''),
             deprecated=doclet.get('deprecated', ''),
-            seealso=doclet.get('see', []),
+            see_also=doclet.get('see', []),
             content='\n'.join(self._content))
 
 
@@ -237,7 +237,7 @@ class AutoClassRenderer(JsRenderer):
             fields=self._fields(doclet),
             examples=doclet.get('examples', ''),
             deprecated=doclet.get('deprecated', ''),
-            seealso=doclet.get('see', []),
+            see_also=doclet.get('see', []),
             class_comment=doclet.get('classdesc', ''),
             constructor_comment=doclet.get('description', ''),
             content='\n'.join(self._content),
@@ -319,7 +319,7 @@ class AutoAttributeRenderer(JsRenderer):
             name=name,
             description=doclet.get('description', ''),
             deprecated=doclet.get('deprecated', ''),
-            seealso=doclet.get('see', []),
+            see_also=doclet.get('see', []),
             examples=doclet.get('examples', ''),
             content='\n'.join(self._content))
 
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -171,10 +171,15 @@ class JsRenderer(object):
         used_names = []
         MARKER = object()
 
-        for name, default, type in [(param['name'].split('.')[0],
-                                     param.get('defaultvalue', MARKER),
-                                     param.get('type', {'names': []}))
-                                    for param in doclet.get('params', [])]:
+        for param in doclet.get('params', []):
+            name = param['name'].split('.')[0]
+            default = param.get('defaultvalue', MARKER)
+            type = param.get('type', {'names': []})
+
+            # Add '*' to the parameter name if it's a variadic argument
+            if param.get('variable'):
+                name = '*' + name
+
             if name not in used_names:
                 params.append(rst.escape(name) if default is MARKER else
                               '%s=%s' % (rst.escape(name),
--- a/sphinx_js/renderers.py
+++ b/sphinx_js/renderers.py
@@ -176,9 +176,9 @@ class JsRenderer(object):
             default = param.get('defaultvalue', MARKER)
             type = param.get('type', {'names': []})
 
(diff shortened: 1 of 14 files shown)


erikrose:
Thanks; great stuff! This is about 3rd in my sphinx-js review queue at the moment.

sphinx_js/templates/function.rst
```diff
@@ -1,5 +1,12 @@
 .. js:function:: {{ name }}{{ params }}
 
+   {% if deprecated -%}
+   .. note::
+
+      This function is deprecated.
```
erikrose: > There's a lot of duplication now between the three directives' templates; I'm wondering if it might be helpful to factor them into a shared ReST template somehow.

If you wanted to factor this stuff out, you could just shorten it to, e.g., "Deprecated. Blah blah blah.", without mention of the kind of construct it is.

erikrose: Btw, I like your choice to render it as a note. I think that gives it the right level of emphasis.

erikrose: Actually, I think the best option would be a colon, since that's how jsdoc renders it by default. Then capital and lowercase letters, as well as complete and incomplete clauses, will read as intended.

"Deprecated: blah blah blah".

sphinx_js/renderers.py
```diff
@@ -171,10 +171,15 @@ def first(list, default):
...
-        for name, default, type in [(param['name'].split('.')[0],
-                                     param.get('defaultvalue', MARKER),
-                                     param.get('type', {'names': []}))
-                                    for param in doclet.get('params', [])]:
+        for param in doclet.get('params', []):
+            name = param['name'].split('.')[0]
+            default = param.get('defaultvalue', MARKER)
+            type = param.get('type', {'names': []})
+
+            # Add '*' to the parameter name if it's a variadic argument
```
erikrose: > I rendered variadics as a Python star-arg, but on second thought, it'd probably make more sense to change it to an ellipsis.

I agree, if Sphinx is willing to render it reasonably.

sphinx_js/renderers.py
```diff
@@ -217,6 +222,8 @@ def _template_vars(self, name, full_path, doclet):
             fields=self._fields(doclet),
             description=doclet.get('description', ''),
             examples=doclet.get('examples', ''),
+            deprecated=doclet.get('deprecated', ''),
+            seealso=doclet.get('see', []),
```
erikrose: `see_also`, for PEP 8 compliance

sphinx_js/templates/attribute.rst
```diff
@@ -15,3 +22,11 @@
    {%- endif %}
 
    {{ content|indent(3) }}
+
+   {% if seealso -%}
+   .. seealso::
```
erikrose: Nice find! I didn't know about this directive.

sphinx_js/templates/attribute.rst
```diff
@@ -15,3 +22,11 @@
...
+
+      {% for reference in seealso -%}
+      - :any:`{{ reference }}`
```
erikrose: Does this work in the js domain? I [see it documented](http://www.sphinx-doc.org/en/stable/domains.html) for C++ but nowhere else.

lidavidm: It seems to be [new to Sphinx 1.3](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-any), and should work with any domain. Do we want to be compatible with earlier versions of Sphinx though?

sphinx_js/renderers.py
lidavidm: I changed it to an ASCII ellipsis; I thought about a Unicode one, but that isn't actually valid JavaScript.

lidavidm:
Alright, thanks for the review! I've factored out the common template bits into Jinja2 macros (ReST includes don't work since they're not relative to the template location, but rather the documentation source file locations; Jinja2 includes don't work since they don't indent properly). If this all looks good I'll get the test cases written up over the next few days.

lidavidm:
I also seem to have broken my own tests, so I'll get that fixed when I can.

sphinx_js/templates/attribute.rst
erikrose: No, I'm not concerned with it.

sphinx_js/templates/common.rst
```diff
@@ -0,0 +1,30 @@
+{% macro deprecated(message) %}
+{% if message -%}
+.. note::
+
+   Deprecated
+   {%- if message is string -%}: {{ message }}{% else %}.{% endif -%}
```
erikrose: Shouldn't this be just `if message`? You default in the `.get()` call is `''`, so it will always be a string. Otherwise, looks great.

lidavidm: Ah what happens is if you just have `@deprecated` in JSDoc, then the doclet contains `deprecated: True`. So we have to check if the field is actually a string before rendering it.

erikrose:
Brilliant. :-)

tests/test_build/test_build.py
```diff
@@ -108,6 +108,15 @@ def test_autofunction_variadic(self):
             '      * **a** --\n\n'
             '      * **args** --\n')
 
+    def test_autofunction_deprecated(self):
+        """Make sure @deprecated tags can be documented with autofunction."""
+        self._file_contents_eq(
+            'autofunction_deprecated',
+            u'deprecatedFunction()\n\n'
```
erikrose: This u must be left over from when you were using the unicode ellipses.

tests/test_build/test_build.py
```diff
@@ -108,6 +108,15 @@ def test_autofunction_variadic(self):
...
+            '   Note: Deprecated.\n\n'
+            'deprecatedExplanatoryFunction()\n\n'
+            '   Note: Deprecated: don\'t use anymore\n')
```
erikrose: If you like, you can use double quotes here; the linter won't complain.

lidavidm:
Thanks! I'm still working on the last set of test cases (currently in the middle of an apartment hunt), I'm hoping to get it done in the next couple days. I'll ping you once it's finished.

lidavidm:
@erikrose all ready!

sphinx_js/renderers.py
```diff
@@ -309,6 +318,8 @@ def _template_vars(self, name, full_path, doclet):
         return dict(
             name=name,
             description=doclet.get('description', ''),
+            deprecated=doclet.get('deprecated', False),
```
erikrose: Just had an idea: someday we should factor out the common kwargs into a function we call from each of the function/class/attr renderers. Then we wouldn't have to repeat the tests 3x each time.

lidavidm: Sphinx itself seems to use a set of mixins (https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/docfields.py), though I haven't looked closely enough to figure out how everything gets arranged there.

erikrose:
Version 2.6 is out and includes this PR. Thanks again!

PR #70, opened 2018-08-05
Support freeform text @see tag

bookmoons:
Proposing support for `@see` tags with freeform text values. Tests are included.

The [JSDoc manual](http://usejsdoc.org/tags-see.html) gives 2 forms of `@see` values: a reference and freeform text. The current logic only supports references. This change adds support for the text form.

```js
/**
 * @see module:frobnicator.Frobnicator#frobnicate
 * @see Linux manpage for foo.
 */
```

The extant support seems to interpret values as references to Sphinx entities rather than JSDoc namepaths. This change keeps that behavior for references.

The new logic interprets the value as text if it contains any space characters. This is a bit of a hack but should work for most cases. If JSDoc namepath support is ever added, it could be changed to attempt namepath resolution and failover to text for consistency with JSDoc behavior.

Prepares for supporting Markdown in `@see` text values.

### DIFF
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -65,6 +65,9 @@ def gather_doclets(app):
             segments = doclet_full_path(d, root_for_relative_paths, longname_field='memberof')
             app._sphinxjs_doclets_by_class[tuple(segments)].append(d)
 
+    # Process @see tags
+    process_see(doclets, app._sphinxjs_doclets_by_path)
+
 
 def program_name_on_this_platform(program):
     """Return the name of the executable file on the current platform, given a
@@ -139,6 +142,23 @@ def analyzer_for(language):
     except KeyError:
         raise SphinxError('Unsupported value of js_language in config: %s' % language)
 
+def process_see(doclets, suffix_tree):
+    """Process @see tags for all doclets."""
+    for doclet in doclets:
+        if 'see' in doclet:
+            see = doclet['see']
+            process_see_list(see, suffix_tree)
+
+def process_see_list(see, suffix_tree):
+    """Process a list of @see tag values.
+
+    Leaves any value with a space as freeform text.
+    Translates other values into RST crossreferences.
+    """
+    for i, item in enumerate(see):
+        if ' ' not in item:
+            item = ':any:`%s`' % item
+            see[i] = item
 
 def root_or_fallback(root_for_relative_paths, abs_source_paths):
     """Return the path that relative JS entity paths in the docs are relative to.
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -161,7 +161,7 @@ def process_see_list(see):
         if ' ' in item:
             item = process_see_item_text(item)
         else:
-            item = ':any:`%s`' % item
+            item = ':any:`%s`' % item.strip()
         see[i] = item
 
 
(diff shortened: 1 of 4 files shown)


bookmoons:
Corrected a couple blemishes here.

Also hit a new case and realized it wasn't handling multiline text values. Updated to add support.

erikrose:
See https://github.com/erikrose/sphinx-js/issues/69#issuecomment-411657301 for an example when reviewing.

sphinx_js/doclets.py
```diff
@@ -140,6 +143,39 @@ def analyzer_for(language):
         raise SphinxError('Unsupported value of js_language in config: %s' % language)
 
 
+def process_see(doclets):
+    """Process @see tags for all doclets."""
+    for doclet in doclets:
+        if 'see' in doclet:
+            see = doclet['see']
+            process_see_list(see)
```
erikrose: Generally I like to be more explicit that the routine is modifying the passed-in data structure. You could do it through naming the routine differently, but an even better way is to do an assignment here, e.g. `doclet['see'] = process_see_list(...)`.

sphinx_js/doclets.py
```diff
@@ -140,6 +143,39 @@ def analyzer_for(language):
...
+
+
+def process_see_list(see):
+    """Process a list of @see tag values.
+
+    Leaves any value with a space as freeform text.
+    Translates other values into RST crossreferences.
+    """
```
erikrose: Most other places in the code, we've been leaving an extra blank line at the end of multiline docstrings. Btw, I loooooove your good docstrings. So many people neglect them, but they are the first place I go to understand new code.

sphinx_js/doclets.py
```diff
@@ -140,6 +143,39 @@ def analyzer_for(language):
...
+    for i, item in enumerate(see):
+        if ' ' in item:
+            item = process_see_item_text(item)
+        else:
+            item = ':any:`%s`' % item.strip()
+        see[i] = item
+
+
+def process_see_item_text(item):
```
erikrose: All the details of your processing look fine. But so far, we've pushed off all the formatting concerns like these to renderers.py and kept doclets.py about unmodified, raw doclets. I'd like to keep that distinction. Do you think you could move this machinery into the renderers module?

Naively, that'll cause some repetition, like of `see_also=process_see(doclet.get('see', []))`, across classes. This might be a good time to factor out some of the other common bits as well, like `examples=doclet.get('examples', '')` and `deprecated=doclet.get('deprecated', False),`. As a bonus, stating all that just once in the code will mean we no longer need to have a Cartesian product of tests, as you suffered from in having to test your `see` functionality 3 times.

sphinx_js/doclets.py
```diff
@@ -140,6 +143,39 @@ def analyzer_for(language):
...
+
+
+def process_see_list(see):
+    """Process a list of @see tag values.
+
+    Leaves any value with a space as freeform text.
+    Translates other values into RST crossreferences.
+    """
+    for i, item in enumerate(see):
+        if ' ' in item:
```
erikrose: I think your heuristic is fine for now. Of course, you could implement the very test you described elsewhere, trying to resolve the path and falling back to freeform on failure. All you have to do is look it up in _sphinxjs_doclets_by_path. Grep for that in renderers.py for an example.

bookmoons:
Thanks very much for this. I'm just seeing it. I'm dug into something for a while here, so will go through all of this as soon as I can.

rhelmer:
@bookmoons this looks super close to landing, I think you could breeze through Erik's review comments. Could you take a few minutes to finish this up? I'd be more than happy to do a final review and merge it for you. Thanks!

bookmoons:
Thanks rhelmer. I probably won't get back to this so I'm closing it. Feel free to scavenge the code.

Issue #73, opened 2018-09-14
autoclass directive cuts off module name when used with a typescript interface


Towerism:
It renders: "interface, exported from odule_her", but it should say "interface, exported from module_here". The first and last letter are cut off from the module name. A quick glance at the code tells me the line referenced below is the culprit. It seems oddly specific to leave off the first and letter, doesn't it? I would guess that earlier versions of typedoc would display quotes around the module name; however, it seems incorrect when the latest version of typedoc is used, as the module name gets cut off.

https://github.com/erikrose/sphinx-js/blob/cb249103ed394f670cd316ce0e9e31424cb3a341/sphinx_js/typedoc.py#L257

erikrose:
@WimYedema Any ideas?

WimYedema:
Indeed, when I look at a typedoc json file I tested this on I see this:

`                        "name": "\"class\"",
                        "kind": 1,
                        "kindString": "External module",
`

We could test for the presence of quotes and remove them if they're there...

Towerism:
I was thinking that was the issue. If you guys could use the extra capacity, I can submit a pull request with this change.

erikrose:
We're always open to more capacity. :-) My only concern here is whether this approach will be a reliable one. It sounds like sometimes there are quotes and sometimes there aren't. It is *possible* for an attr name to start with a quote, e.g.

```js
{'"foo"': 'weird attr name'}
```

If we can figure out a reliable approach, that'd be preferable. But we've been forced to guess by jsdoc before, so guessing right most of the time is better than acting wrong.

Towerism:
Is cleansing the name really desirable then? Maybe it would be better to just "do whatever the typedoc/jsdoc" in this case.

erikrose:
I generally try to stick to the other tools’ behavior if I can’t do something clearly better. That way, a codebase isn’t coupled to any tool. Want to take a look and see what typedoc generates on its own?

erikrose:
What I meant was "What does the formatted (HTML) documentation that typedoc emits look like?" Are they also having spurious quotes?

Towerism:
Oh no, no spurious quotes in the formatted document emitted by typedoc; that plus the JSON being free of spurious quotes leads me to believe it's a version difference.

erikrose:
Sounds like stripping a single surrounding pair of " or ' is the best approach.

erikrose:
[...]

PR #74, opened 2018-09-16
Strip quotes with string.strip instead of [1:-1]

Towerism:
Moving the conversation from #73, so we can have code to look at and discuss. This change maintains the existing intended functionality of removing quotes from module name. However, with the change only the quotes are stripped. If there are no quotes to begin with, the original name is left in-tact.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -100,9 +100,9 @@ class TypeDoc(object):
             else:
                 longname += '~'
         if kindString == 'Module':
-            return longname + 'module:' + node.get('name')[1:-1]
+            return longname + 'module:' + node.get('name').strip("'\"")
         elif kindString == 'External module':
-            return longname + 'external:' + node.get('name')[1:-1]
+            return longname + 'external:' + node.get('name').strip("'\"")
         else:
             return longname + node.get('name')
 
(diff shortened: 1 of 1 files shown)


erikrose:
I like the idea of stripping any initial quote or double-quote. It at least improves things over just lopping off any char that happens to be there. However, strip() will strip any number of leading or trailing quotes, not stopping at one.

Towerism:
Well, we're talking about module names, so I think the code doesn't need to be that so elaborate as to limit the strip to just one set of quotes.

erikrose:
Weird things happen. :-) Filenames can have quotes in them, just as attributes can.

Towerism:
Ah. Well then I think it would be best to leave whatever quotes there may be as they are since there is no way to know whether or not they should be deleted.

graup:
I still think we should change the code, always stripping the first and last character seems like a code smell. I guess [something like this](https://stackoverflow.com/a/20577580) may be the proper solution?

erikrose:
I got rid of the "exported from" link in the big TS rewrite of #120. Does anybody miss it? If so, most of the hard work toward showing it is already in there. Please open a ticket for that if so.

I also rewrote a lot of the TS analysis, though the slicing survives in ones place. I'd be interested to see if the bug described here is still reproducible, if we were to print the "exported from" bits.

PR #79, opened 2018-11-15
Use confdir as jsdoc's runtime directory

### LINKED ISSUE #78: jsdoc working directory
Hi and thanks for sphinx-js

When specifying a jsdoc.json file which has excluded directory defined by a relative path, these directories are evaluated from the directory jsdoc is run from, i.e. the directory sphinx-build is run from. This directory may change depending on configurations, and it would therefore be interesting to be able:

- either to specify jsdoc's working directory
- force jsdoc to be run in the conf.py file's directory

The change is relatively trivial (passing a working directory to subprocess.Popen) I'll be happy to create a PR after discussion on what would the best solution be.

tkhyn:
This resolves #78


erikrose:
Thanks much!

PR #81, opened 2018-11-22
Failing test case showing messed up order of constructor

graup:
Output of this test

```
======================================================================
FAIL: test_autoclass_constructor (test_build.Tests)
Make sure constructor comes before methods.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/paulgrau/projects/opensource/sphinx-js/tests/test_build_ts/test_build.py", line 17, in test_autoclass_constructor
    assert_true(pos_method > pos_cstrct, "Constructor appears after method in " + contents)
AssertionError: Constructor appears after method in class ClassDefinition()

   A definition of a class

   ClassDefinition.method1(simple)

      This is a method without return type

      Arguments:
         * **simple** (*number*) -- A parameter with a simple type

   ClassDefinition.new ClassDefinition(simple)

      ClassDefinition constructor

      Arguments:
         * **simple** (*number*) -- A parameter with a simple type

      Returns:
         **class.ClassDefinition** --


----------------------------------------------------------------------
```

Expected build output (correct order):

```
class ClassDefinition()

   A definition of a class

   ClassDefinition.new ClassDefinition(simple)

      ClassDefinition constructor

      Arguments:
         * **simple** (*number*) -- A parameter with a simple type

      Returns:
         **class.ClassDefinition** --

   ClassDefinition.method1(simple)

      This is a method without return type

      Arguments:
         * **simple** (*number*) -- A parameter with a simple type
```

but the `ClassDefinition.new ClassDefinition(simple)` is a bit ugly, too. Not sure what the best way would be.

### DIFF
--- a/tests/test_build_ts/source/class.ts
+++ b/tests/test_build_ts/source/class.ts
@@ -1,0 +1,20 @@
+/**
+ * A definition of a class
+ */
+class ClassDefinition {
+    /**
+     * ClassDefinition constructor
+     * @param simple A parameter with a simple type
+     */
+    constructor(simple: number) {
+
+    }
+
+    /**
+     * This is a method without return type
+     * @param simple A parameter with a simple type
+     */
+    method1(simple: number) : void {
+
+    }
+}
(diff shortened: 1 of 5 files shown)


erikrose:
Does #99 obsolete this?

graup:
Yep, thanks.

PR #91, opened 2019-01-15
Fix typedoc parsing code broken since typedoc@0.14.0. Fixes #90

graup:
Without this change and typedoc@0.14.0, the tests of sphinx-js actually fail (same error as reported in #90). This change makes tests pass again. 

I *think* there will always be only one signature per accessor? But not 100% sure about that.

## Steps to reproduce

    npm install -g typedoc@0.14.0
    python setup.py test

Fails on current master, passes with this change.

PR #92, opened 2019-01-21
Fix issue 82

### LINKED ISSUE #82: typedoc order of class methods with autoclass?
When using autoclass in javascript mode, the constructor documentation always appears at the top of the class. This is messed up in typescript mode.

It seems that all class instance methods are ordered alphabetically, and in typescript mode the constructor ends up being a method called "new Class()" which is sorted under N (and not at the top as it used to be in normal js mode). 

I created a failing test case showing the problem: https://github.com/erikrose/sphinx-js/pull/81

s-weigand:
This PR fixes following issues when using Typescript:
1. That the constructor of a class was listed as a separate method with name `new <ClassName>`, now the docs the Class and the constructor are combinded and the arguments are shown in the class head.
2. Union types for arguments now get properly rendered and not split.

The changes in the code basically are, so the typedoc class doclet mimics the jsdoc one.
That for if `kindstring` is 'class' it will look up the relevant information in `Constructor signature`, combine it with the ones already provided and write it to the class. Since I didn't know in which cases 'isAbstract' and 'isOptional', I just copied the code to the other lookup.
For the union types to work properly `TypeDoc.make_type` needs to return a list of strings, not a list with one string joined on `|`, maybe this also applies to other types (`complex`?)

Here are two screenshots, so you can see the difference.
**Rendered version with sphinx-js 2.7.1:**
be4_img_edited

**Rendered version with my fix in place:**
after_img_edited

**Closing issues**
closes #82


graup:
Hi, thank you for this. I'm currently struggling with the same thing. I'll review your changes and report back soon.

sphinx_js/typedoc.py
```diff
@@ -138,7 +138,7 @@ def make_type_name(self, type):
             names = ['[' + ','.join(types) + ']']
         elif type.get('type') == 'union':
             types = [self.make_type_name(t) for t in type.get('types')]
-            names = [' | '.join(types)]
+            return types
```
graup: I think this whole method is a bit weird, why make single item lists and then convert them back to a string... Do you want to refactor it? If not I can give it a try in another PR.

sphinx_js/typedoc.py
```diff
@@ -153,9 +153,15 @@ def make_type_name(self, type):
...
-        }
+        type_name = self.make_type_name(type)
+        if isinstance(type_name, list):
+            return {
+                'names': type_name
+            }
+        else:
+            return {
+                'names': [type_name]
+            }
```
graup: Same as previous comment, if we refactor make_type_name this wouldn't be necessary, would it?

sphinx_js/typedoc.py
```diff
@@ -167,19 +173,27 @@ def make_description(self, comment):
...
         """Construct a jsdoc parameter entry"""
         typeEntry = param.get('type')
+        description = self.make_description(param.get('comment'))
+        name = param.get('name')
+        if parent_comment is not None and parent_comment.get('tags') is not None:
+            for tag in parent_comment.get('tags'):
+                tag_text = tag.get('text')
+                if tag.get('tag') == 'arg' and tag_text.startswith(name):
+                    tag_text.replace(name, '', 1)
+                    description = tag_text
```
graup: Could you add a comment to the this if block? IMO it's not immediately obvious why it is needed.

graup:
Also, could you rebase this on master? Without the already merged fix from #91 this doesn't actually work.

sphinx_js/typedoc.py
s-weigand: Should be done with commit 31c4aba, please have a look again.

sphinx_js/typedoc.py
s-weigand: Should be done with commit 31c4aba, please have a look again.

sphinx_js/typedoc.py
s-weigand: Sometimes `typedoc` writes arguments to `tags` of the parent, and this code extracts them.
I have an example at my work PC and will have a look at it + implement a test and comments toworrow.

s-weigand:
While refactoring `make_type` and `make_type_name`, I also fixed a problem with generic types and `keyof` generic type.
spinx-js-generic-rendering

erikrose:
@graup When it looks good to both you and the PR author, I'll merge it. I don't use TypeScript, so I'm happy to use the buddy system here. :-) (Once this is merged, I will, as you suggest, have a look at #99.)

sphinx_js/typedoc.py
s-weigand: This code extracted the parameter docstring if instead of the keyword `@param`, the aliases `@arg` or `@argument` were used.

s-weigand:
I also implemented some code that extracts the parameter description if someone uses an alias of `@param`, namely `@arg` and `@argument` ([see](https://github.com/Microsoft/TypeScript/wiki/JsDoc-support-in-JavaScript)). Those are treated diffrently by `typedoc` and saved in `comment.tags` of the `params` parent node (methods)/ parents child node (constructor).
Ofc I also wrote a testcase for this.

### Old rendering
alias_render_old

### New rendering
alias_render_new

graup:
I tested this and it works for me. Tests look good too, so I'd say let's get this merged.

I'll push an update to #99 later.

s-weigand:
Should I rename the `*.jsdoc.ref` files to `*.jsdoc` before merge?

erikrose:
If I understand properly, then yes. It looks like you changed the output format a bit and named those "ref" temporarily, correct?

s-weigand:
I just followed the comment in `tests/test_typedoc.py test_references`

```
# When a reference file is missing, this is probably a new test.
# Generate a reference file for the developer to review.  If the
# generated reference is good, just remove the '.ref' extension.
```

erikrose:
A great many thanks for the fix and apologies for the delays!

PR #93, opened 2019-01-21
Change url in setup.py from a fork to the main repo

s-weigand:
When I initially found this project on PyPi I followed the link to the homepage and got to the fork of @erikrose and had to click again to get here. Even so this isn't a big issue, it is always nice to save a click here and there 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,3 @@ sphinx_js.egg-info/
 */__pycache__/*
 # Python 2.7
 *.pyc
-# Pycharm config
-.idea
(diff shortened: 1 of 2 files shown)


s-weigand:
That PyCharm commit did sneak itself on my local master and to prevent a merge conflict with PR #92 i removed it again.

erikrose:
Originally I had sphinx-js under my own private account, then I moved it to the Mozilla org. Thanks for cleaning up the residual weirdness! :-)

Issue #94, opened 2019-02-05
Exception Occurred: AttributeError 'Value'  object has no attribute 'keys'


timkpaine:
Error comes from [this line](https://github.com/mozilla/sphinx-js/blob/master/sphinx_js/suffix_tree.py#L48). 

<img width="892" alt="screen shot 2019-02-05 at 12 18 06 pm" src="https://user-images.githubusercontent.com/3105306/52291461-66bfc380-2940-11e9-97cb-89b008890914.png">

Showing that the named tuple has no attribute keys: [from here](https://github.com/timkpaine/sphinx-js/blob/master/sphinx_js/suffix_tree.py#L6).

Was it meant to be `next(iterkeys(tree.values))`? Will try to patch on my fork. 

sphinx == 1.7.2
sphinx-js == 2.7.1
six == 1.12.0

timkpaine:
closing from my queue

PR #95, opened 2019-02-05
fix for #94

timkpaine:
Fixes bug where tree can be an instance of Value, and therefore not utterable #94


erikrose:
Thank you! Looking forward to reviewing this, though I'll have to load the project back into my head. If you could provide a reduced test case or (this would be amazing) a new test in test_suffix_tree.py, it would speed up acceptance.

timkpaine:
Will do

Issue #98, opened 2019-03-21
typedoc: generic types


graup:
At the moment, generic types such as `Promise<Something>` or `Partial<Something>` only show the first part in the generated documentation which is pretty useless.

We should at least support showing more details on common ones such as Promise, Partial, Array etc., but ideally make it work for any kind of generic.

I might get a chance to look into implementing this myself soon-ish. But if someone else wants to take a stab at it, let me know.

PR #99, opened 2019-03-22
Typescript doclet order and generic types

graup:
This is a fix for #98. It builds on top of #92 (should be merged first). This works by recursively calling `make_type_name` when we have `typeArguments`.

This PR additionally changes the order of fields (they should come before methods) by using the `longname` for sorting instead of just the name.

I also added a build test for typescript.

Let's merge #92 first and then review this PR separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ sphinx_js.egg-info/
 .idea
 # VsCode config
 .vscode
+.DS_Store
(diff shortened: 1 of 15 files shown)


graup:
I rebased this on the current master and cleaned up the commit history.

Summary of changes:
- renderer: use longname to sort, if available
- add a typescript build test to test the order of doclets in the rendered output
- doclet generator: support type arguments
- add ts->jsdoc tests for complex type arguments

graup:
@erikrose Here's a gentle reminder to look at this PR when you've got some time. Thanks.

PR #111, opened 2019-08-21
doclets: Allow overloaded functions to not conflict

### LINKED ISSUE #110: TypeScript overloads fail
The following example:

```
function foo(bar: number): number;
function foo(bar: string): string;

function foo(bar: number | string): number | string {
    return bar;
}
```

Generates this error:

```
Exception occurred:
  File ".../site-packages/sphinx_js/doclets.py", line 49, in gather_doclets
    raise PathsTaken(conflicts)
sphinx_js.doclets.PathsTaken: Your JS code contains multiple documented objects at each of these paths:

  ./docs/foo.external:foo.foo

We won't know which one you're talking about. Using JSDoc tags like @class might help you differentiate them.
```

tavianator:
Fixes #110.

### DIFF
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -45,7 +45,9 @@ def gather_doclets(app):
                 doclet_full_path(d, root_for_relative_paths),
                 d)
         except PathTaken as conflict:
-            conflicts.append(conflict.segments)
+            # Allow overloaded functions, but not other duplicates
+            if d['kind'] != 'function':
+                conflicts.append(conflict.segments)
     if conflicts:
         raise PathsTaken(conflicts)
 
--- a/tests/test_build_ts/source/class.ts
+++ b/tests/test_build_ts/source/class.ts
@@ -26,4 +26,18 @@ class ClassDefinition {
     anotherMethod() {
 
     }
+
+    /**
+     * This is an overloaded method (1).
+     */
+    overloaded(arg: number): number;
+
+    /**
+     * This is an overloaded method (2).
+     */
+    overloaded(arg: string): string;
+
+    overloaded(arg: number | string): number | string {
+        return arg;
+    }
 }
--- a/tests/test_build_ts/test_build_ts.py
+++ b/tests/test_build_ts/test_build_ts.py
@@ -27,3 +27,10 @@ class Tests(SphinxBuildTestCase):
         pos_field = contents.index('ClassDefinition.field')
         pos_method2 = contents.index('ClassDefinition.anotherMethod')
         assert pos_method < pos_field < pos_method2, 'Methods and fields are not in right order in ' + contents
+
+    def test_overloads(self):
+        """Make sure overloaded methods work."""
+        contents = self._file_contents('index')
+        pos_overloaded_1 = contents.index('overloaded method (1)')
+        pos_overloaded_2 = contents.index('overloaded method (2)')
+        assert pos_overloaded_1 < pos_overloaded_2, 'Method overloads are not in right order in ' + contents


erikrose:
1. This is a significant enough change that it requires new tests to demonstrate its efficacy.
2. If there are multiple functions at the same pathname, shouldn't we, at minimum, favor the documented one, if any? It looks like the error you were experiencing happened only when trying to actually use the docs; correct me if I'm wrong. (Of course, the ultimate would be to be able to somehow specify which instance of the function we wanted, but that would need much more design work.)

erikrose:
I forgot to say it, but thanks for the patch! :-) I don't use TS myself, and nobody's maintaining sphinx-js's support at the moment, so I'm very keen to have your interest.

tavianator:
Sure, I'm happy to add some tests for it.

The error I was getting happens when building the docs.  With the new behaviour, both overloads are listed in the docs, which is probably what you want (TypeDoc also lists both, but only one at a time, like this: https://typedoc.org/api/classes/application.html#generatedocs).  Here's what the result looks like in sphinx-js: https://bistring.readthedocs.io/en/latest/JavaScript/Alignment.html#Alignment.originalBounds

The tree that reports the conflicts is used to resolve links I believe.  So when you write something like ``:js:meth:`Alignment.originalBounds` ``, it has to resolve that to the actual element to link to.  This change makes it just pick the first overload arbitrarily.  I'm open to other behaviours here, or other ways of rendering overloads.

You're welcome for the patch! :)

graup:
Tested it quickly, seems to work. This is definitely much better than throwing errors. Let's get some tests for this and merge it!

tavianator:
@graup Added a simple test!

tavianator:
@erikrose The code I changed doesn't distinguish between vanilla JS and TypeScript, so I'm pretty sure it won't.  There's another issue as well: if you do something like `.. js:autofunction:: overloaded` it will only render one of them.  I guess the tree needs to hold a list of matching elements rather than just one.

erikrose:
Hmm, so apparently we have more thinking to do here.

I'll leave it to you and weigh in once you come up with a design you feel good about. Feel free to let config flags be a part of it if necessary; that's fine with me as long as all nontrivial combinations of state have good test coverage.

I have regrets about the state of TS support in sphinx-js. The original contributor gave us a large piece of useful, well-tested code, but it hooked up TS by converting typedoc's output into the ill-specified, oft-changing JSON format `jsdoc -X` emits, because that's what sphinx-js operated on directly at the time (and still does). Had we talked during the design process, I'd have suggested we define a stable intermediate representation and write everything to that. But I was a teensy bit lazy in the 1.0 days, and now it's one of those temporary-begets-permanent situations. I'd still be happy to rewrite the JS parts to hook up to an IR—I think it would be borderline trivial—but I'm not prepared to tackle TS, as typedoc.py is full of branches and a mystery to me. Applicants welcome. :-)

graup:
@erikrose I'd be happy to contribute. Should we open a new issue to design an IR and coordinate working on a branch together?

There are also some problems with some other ts features currently not supported by the jsdoc representation that we could fix then.

tavianator:
I think an IR would be nice.  Would enable us to have TS-specific things in the documentation as first-class citizens.

To avoid needing to mess with typedoc.py at first, since we already have TypeDoc -> JSDoc, we could start with just doing TypeDoc -> JSDoc -> IR.  Then later rewrite the conversion to just do TypeDoc -> IR, and compare the results on some corpus of existing TypeScript projects.

I don't have a tremendous amount of time to contribute to this but I'm happy to review designs and implementations, and maybe do some of the TS side of it.

erikrose:
Alright, I started doing some of the initial spec work toward an IR over at #120. Help is welcome!

erikrose:
It doesn't crash anymore as of the great IR merge of #120, demonstrated by https://github.com/mozilla/sphinx-js/blob/705659ceb08de1ec1fd16bdefd53e7b2e0a8d848/tests/test_typedoc_analysis/test_typedoc_analysis.py#L336-L344. But we should probably show multiple signatures if the Sphinx templates will let us. I've opened #153 to represent that work.

PR #115, opened 2019-09-08
feat: Cache JSDoc results

ptbrowne:
I am working on a documentation site pulling several repositories.

JSDoc data extraction can take 30s. I thought it could help performance
to save the result to a file configured in `app.config`.

EDIT: Build is red, I am looking into it. I welcome any feedback on this PR, I may have pushed the PR a bit early but I did not want it to be lost. Thanks for sphinxjs.

### DIFF
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -1,7 +1,8 @@
 from codecs import getwriter
 from collections import defaultdict
 from errno import ENOENT
-from json import load
+import functools
+from json import load, dump
 import os
 from os.path import abspath, relpath, splitext, sep
 import subprocess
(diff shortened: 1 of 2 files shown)


sphinx_js/doclets.py
```diff
@@ -84,6 +86,29 @@ def make(self):
...
+    def decorator(fn):
+        @functools.wraps(fn)
+        def decorated(*args, **kwargs):
+            filename = get_filename(*args, **kwargs)
+            if filename and os.path.isfile(filename):
+                with open(filename) as f:
+                    return load(f)
+            res = fn(*args, **kwargs)
+            if filename:
+                with open(filename, 'w+') as f:
```
erikrose: It shouldn't technically matter since you arrive here only if the file doesn't exist, but why not use the more common "w" mode rather than the appending "w+"?

ptbrowne: If I am not mistaken, the "+" is to automatically create the file if it does not exist, the "a" mode is for appending.

ptbrowne:
I haven't had time to work again on this PR, sorry for the delay.

erikrose:
My only big concern is this: how do you ensure the cache file is removed when stale? You may have a nice build process that ensures that happens when the JS code changes, but I doubt most potential sphinx-js users have that set up. And it would probably be nontrivial—though possible—to have sphinx-js itself handle that.

ptbrowne:
I understand you concern, I had the same worry, but then I thought that since it is opt-in, a user wanting to use this would only use this knowing that it should remove the cache manually if thinking it is stale (by `rm`ing it).

sphinx_js/doclets.py
erikrose: Looks like we were both wrong! "+" opens the file for reading *and* writing. "a" is indeed for appending. "w" is for truncating. So I think "w" is the best fit here.

ptbrowne: Thanks, looks like we are not the only ones being confused : https://stackoverflow.com/questions/16208206/confused-by-python-file-mode-w/16208298, I am updating the PR.

erikrose:
Ah, I missed that it was opt-in. Of course it is. Part of this PR should be adding docs saying as much (and explaining how to use it, including an admonition to clean up your stale caches) to the readme under Configuration Reference.

erikrose:
When you come back to this, you might want to merge in from master. We just switched test runners, which might make your debugging easier.

erikrose:
Merged! Thanks for the patch! If you want to be sure it doesn't accidentally break later, you should also write up some tests that exercise it. Thanks again!

Issue #109, opened 2019-07-04
Rule 'path_and_formal_params' didn't match at


joshglenn:
I am using sphinx-js for the first time. When using this directive:

```
.. js:autoclass:: MyClass
  :members:
```

I get successful output from my test class called MyClass.

However, several of my modules have identical names. So I wanted to test the autoclass directive telling the filename directly, like this:

```
.. js:autoclass:: ../mson-react/MyClass.js.MyClass
  :members:
```

I get an error message:

```
Exception occurred:
  File "/Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/parsimonious/expressions.py", line 127, in match
    raise error
parsimonious.exceptions.ParseError: Rule 'path_and_formal_params' didn't match at '../mson-react/MyClas' (line 1, column 1).
```

Any idea how I can diagnose this?

joshglenn:
Here is the traceback:

```
# Sphinx version: 1.8.5
# Python version: 3.7.1 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10.1
# Last messages:
#   reading sources... [ 14%] rst_cheatsheet
#   
#   reading sources... [ 28%] mson-react
#   
#   reading sources... [ 42%] mson
#   
#   reading sources... [ 57%] custom-components
#   
#   reading sources... [ 71%] index
#   
# Loaded extensions:
#   sphinx.ext.mathjax (1.8.5) from /Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/sphinx/ext/mathjax.py
#   alabaster (0.7.12) from /Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/alabaster/__init__.py
#   sphinx_js (unknown version) from /Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/sphinx_js/__init__.py
Traceback (most recent call last):
  File "/Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/Users/josh/.virtualenvs/mson-docs/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
[...]

PR #116, opened 2019-09-08
feat: Be more lenient on errors if app.config.sphinxjs_lax is set

ptbrowne:
When integrating sphinxjs for a multi repository documentation site, I stumbled
upon several conflicts errors and `Rule 'path_and_formal_params' didn't match at` (#109).

For my usecase, it is more convenient to ignore those errors and fix them
on a case by case basis, so I do not mind if I ignore those doclets.

This PR adds a config attribute `sphinxjs_lax` that make sphinxjs doclet generation
not fail if it has errors.

### DIFF
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -8,12 +8,16 @@ from os.path import abspath, relpath, splitext, sep
 import subprocess
 from tempfile import TemporaryFile, NamedTemporaryFile
 
+from parsimonious.exceptions import ParseError
 from six import string_types
 from sphinx.errors import SphinxError
+from sphinx.util.logging import getLogger
 
 from .parsers import path_and_formal_params, PathVisitor
 from .suffix_tree import PathTaken, SuffixTree
 from .typedoc import parse_typedoc
+
+logger = getLogger(__name__)
 
 
 def gather_doclets(app):
@@ -46,8 +50,17 @@ def gather_doclets(app):
                 d)
         except PathTaken as conflict:
             conflicts.append(conflict.segments)
+        except ParseError:
+            if not app.config.sphinx_js_lax:
+                raise
+            else:
+                logger.warning('Could not parse correctly %s' % d)
     if conflicts:
-        raise PathsTaken(conflicts)
+        exception = PathsTaken(conflicts)
+        if not app.config.sphinx_js_lax:
+            raise exception
+        else:
+            logger.warning('%s' % exception)
 
     # Build lookup table for autoclass's :members: option. This will also
     # pick up members of functions (inner variables), but it will instantly
--- a/sphinx_js/doclets.py
+++ b/sphinx_js/doclets.py
@@ -50,11 +50,11 @@ def gather_doclets(app):
                 d)
         except PathTaken as conflict:
             conflicts.append(conflict.segments)
-        except ParseError:
+        except ParseError as e:
             if not app.config.sphinx_js_lax:
                 raise
             else:
-                logger.warning('Could not parse correctly %s' % d)
+                logger.warning('Could not parse path correctly %s' % e.text)
     if conflicts:
         exception = PathsTaken(conflicts)
         if not app.config.sphinx_js_lax:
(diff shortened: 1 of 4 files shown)


sphinx_js/doclets.py
```diff
@@ -45,8 +48,16 @@ def gather_doclets(app):
                 d)
         except PathTaken as conflict:
             conflicts.append(conflict.segments)
+        except parsimonious.exceptions.ParseError as e:
+            if not app.config.sphinx_js_lax:
+                raise
+            else:
+                print('Could not parse correctly %s' % d)
```
erikrose: This PR seems reasonable, but let's use Sphinx's logging API rather than print(): https://www.sphinx-doc.org/en/master/extdev/logging.html

sphinx_js/doclets.py
```diff
@@ -45,8 +48,16 @@ def gather_doclets(app):
                 d)
         except PathTaken as conflict:
             conflicts.append(conflict.segments)
+        except parsimonious.exceptions.ParseError as e:
```
erikrose: And let's import this deeply for consistency with the rest of the project. Also, `e` isn't used, so we can elide it.

erikrose (changes requested):
Good idea for big projects. Thanks for the patch!

sphinx_js/doclets.py
ptbrowne: Import this deeply ?

ptbrowne: Ah, I guess I should do `from parsimonious.exceptions import ParseError` then.

sphinx_js/doclets.py
```diff
@@ -3,17 +3,20 @@
...
 
 from six import string_types
 from sphinx.errors import SphinxError
+from sphinx.util import logging
 
 from .parsers import path_and_formal_params, PathVisitor
 from .suffix_tree import PathTaken, SuffixTree
 from .typedoc import parse_typedoc
 
+logger = logging.getLogger(__name__)
```
erikrose: Might as well import more deeply. "logging" doesn't add any useful context to "getLogger"; it's just more to read.

sphinx_js/doclets.py
```diff
@@ -45,8 +48,17 @@ def gather_doclets(app):
                 d)
         except PathTaken as conflict:
             conflicts.append(conflict.segments)
+        except ParseError:
+            if not app.config.sphinx_js_lax:
+                raise
+            else:
+                logger.warning('Could not parse correctly %s' % d)
```
erikrose: I'd like to have tests for this output, since formatting a rich datatype like an entire doclet can be error-prone.

sphinx_js/doclets.py
```diff
@@ -45,8 +48,17 @@ def gather_doclets(app):
...
     if conflicts:
-        raise PathsTaken(conflicts)
+        if not app.config.sphinx_js_lax:
+            raise PathsTaken(conflicts)
+        else:
+            for conflict in conflicts:
+                logger.warning('Doclets conflict:%s' % conflict)
```
erikrose: It might be a little nicer to omit the "for" loop and instead print "Doclets conflict:" only once, then format ``conflicts`` afterward nicely.

erikrose (changes requested):

sphinx_js/doclets.py
ptbrowne: I added tests and since the formatting was already done in PathsTaken, I piggy backed on it ;)

sphinx_js/doclets.py
ptbrowne: Done since I use PathsTaken now.

sphinx_js/doclets.py
ptbrowne: Ah yes, I am not used to import deeply like that.

ptbrowne:
Should be OK now, thanks for the review.

sphinx_js/doclets.py
erikrose: You've added tests for conflict errors, but I was talking about parse errors. :-)

erikrose: You've added tests for conflict errors, but I was talking about parse errors. :-)

erikrose (changes requested):

sphinx_js/doclets.py
ptbrowne: I added tests for parse errors :)

erikrose (changes requested):

willkg:

PR #117, opened 2019-09-16
Make tox the unified entrypoint for tests, both local and on CI.

erikrose:
This way, the versions of jsdoc and typdoc we're testing against are consistent, and we don't have to document so much.


erikrose:
Landed in 52f33308dd3491105f46081ce476ecc7ba7b1eaf.

PR #123, opened 2019-11-15
Add LICENSE to distributions

bollwyvl:
Thanks for the great tool!

This adds the LICENSE to the tarball and wheel outputs.

Motivation: looking to package this for conda-forge (which requires licenses be included when the license in question demands it), which can neatly take care of the `nodejs` dependency, and potentially typedoc and jsdoc (though neatly packaging npm stuff is a bit more troublesome). I am having a bit of trouble with the `test_build` stuff, which dies with the `RuntimeError` under the versions available in conda without much feedback from the underlying sphinx call... still investigating!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
 recursive-include sphinx_js/templates *.rst
-
+include LICENSE
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [bdist_wheel]
 universal = 1
+
+[metadata]
+license_files = LICENSE


bollwyvl:
A ha! https://github.com/jsdoc/jsdoc/issues/1328#issuecomment-282143743

> one of the source file's parent directories starts with an underscore. By default, JSDoc ignores source files in directories that begin with an underscore. To change this behavior, replace source.excludePattern with an empty string in your configuration file.

This is getting me because the test directories conda creates during build always have an underscore in their name, so it's automatically filtering them out, even when specified directly. This seems like a particularly nasty gotcha for potential users. Looking into an external fix, but this seems like it could catch users unaware.

erikrose:
Thanks, and thanks for packaging sphinx-js! :-D

bollwyvl:
Awesome, thank you!

Issue #119, opened 2019-10-02
Typescript: generic <A extends B> fails


graup:
Culprit line: https://github.com/mozilla/sphinx-js/blob/2158457eeeafc0baa7aa86033dbe3d9e314223f6/sphinx_js/typedoc.py#L152

`self.make_type_name(arg)` for this case returns `['A', 'extends', ['B']]` which cannot be joined. This line probably needs recursion.

graup:
I fixed it in my fork: https://github.com/graup/sphinx-js/commit/4def189dc364f9c5a24331ae4dc7a4181abb75d5. Would need to add proper tests to merge it back here.

erikrose:
This is fixed in the great IR merge of #120, demonstrated by https://github.com/mozilla/sphinx-js/blob/705659ceb08de1ec1fd16bdefd53e7b2e0a8d848/tests/test_typedoc_analysis/test_typedoc_analysis.py#L375-L379. Thanks!

Issue #120, opened 2019-10-07
Introduce an intermediate representation


erikrose:
Currently, we program everything directly against JSDoc's JSON output. Even the TypeScript stuff first converts to JSDoc. But JSDoc's JSON is unspecified, idiosyncratic (i.e. https://github.com/mozilla/sphinx-js/blob/2158457eeeafc0baa7aa86033dbe3d9e314223f6/sphinx_js/doclets.py#L31-L35), and not always the clearest thing in the world to read at callsites.

Let's instead write against an intermediate representation that we control.

Advantages:
1. We can fix any changes JSDoc makes in one place instead of all over the codebase.
2. We can support information that JSDoc doesn't, as for TypeScript or other frontend languages.
3. Our TypeScript module should become a lot more maintainable.
4. Maybe tests will become easier to write.

What bits of info will we need to support in the IR?
* filename
* file path
* longname (to do: define what this is)
* line number (for error reporting)
* params, with their types, names, default values, descriptions, and whether they're variadic
* exceptions a function can raise
* function return type and description
* examples for functions and classes
* whether funcs or classes are deprecated
* see-alsos
* class comments and class-constructor comments (Should constructor comments be special? Probably; I think we merge them currently.)
* members of a class
  * what kind of thing they are (fuction, typedef (currently a hack, I believe, around not having a proper representation of typedefs—we just shove them through the fuction renderer))
  * access (public, private, protected)
* properties (Is this a TS thing?)

That's from my reading of the JS-related source along with a peek at the (handy) docs at the top of typedoc.py about what data are used. Anybody see anything else? The next step is to gin up a class-based API that supports all this.

graup:
Sounds good so far. We should strive to make this as generic as possible so that we can add new features more easily in the future.

Types are kind of a pain. They need to be nestable and have properties. Perhaps something like
```
functionTypes = [
 {
    name: 'Promise',
    properties: {
        readonly: true,
    },
    children: [
        {
            name: 'TypeA',
        },
   ],
},
...
]
```

There are also keywords like`extends`, and defaults for generic types, and probably more that I haven't encountered yet. I wonder what's the best way to maintain this, as ts will probably add additional syntax features in the future. Typedoc has a [big collection of tests](https://github.com/TypeStrong/typedoc/tree/master/src/test/converter), maybe we can use that.

> properties (Is this a TS thing?)

It is, so far, but it's also a stage3 proposal for JS itself: https://github.com/tc39/proposal-class-fields

[...]

PR #124, opened 2019-11-16
extend typename of typeParameter with constraint

bollwyvl:
> Dupe of #119, probably: feel free to close!

Here's [a fun one](https://github.com/jupyterlab/jupyterlab/blob/bb21bc71085413e14044cdf8aabd08f3126fc76f/packages/application/src/frontend.ts#L185)... yields:

```
# Sphinx version: 2.2.1
# Python version: 3.7.3 (CPython)
# Docutils version: 0.15.2 release
# Jinja2 version: 2.10.3

# typedoc: 0.15.2  # added by me

# Last messages:

Exception occurred:
  File ".../sphinx_js/typedoc.py", line 152, in <listcomp>
    argNames = ['|'.join(self.make_type_name(arg)) for arg in type.get('typeArguments')]
TypeError: sequence item 2: expected str instance, list found
```
(nothing much else in the log)

it appears that `make_type_name` is always expected to return a `List[Text]`: this change ensures a `typeParameter` with a constraint does not become `List[Union[Text, List[Text]]`... or worse! I see some of them are having the `0`th item taken, would that be more appropriate?

Happy to work up a test... i see there are some ts files that are getting round-tripped, which includes a pretty gnarly generic, as well as a build example... what would be most beneficial to add?

However, I can't claim to have gotten all the way down the rabbit hole due to our use of lerna, and trying to make [typedoc-plugin-lerna-packages](https://github.com/marcj/typedoc-plugin-lerna-packages) work as well... my environment is kind of a mess of hacks right now.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -145,7 +145,7 @@ class TypeDoc(object):
             names = [type.get('name')]
             constraint = type.get('constraint')
             if constraint is not None:
-                names.extend(['extends', self.make_type_name(constraint)])
+                names.extend(['extends'] + self.make_type_name(constraint))
         elif type.get('type') == 'reflection':
             names = ['<TODO>']
         if type.get('typeArguments'):


erikrose:
Thank you for the patch! I don't use TypeScript myself, so I have little idea what's going on in our TS support and am thus a poor judge. So I would put it to you: how do you think your fix compares with graup's fix, linked from https://github.com/mozilla/sphinx-js/issues/119#issuecomment-537357793? I am inclined to wait for a test to merge, as well, especially in the absence of better understanding of the code on my part. :-)

erikrose:
This is fixed in the great IR merge of #120, demonstrated by https://github.com/mozilla/sphinx-js/blob/705659ceb08de1ec1fd16bdefd53e7b2e0a8d848/tests/test_typedoc_analysis/test_typedoc_analysis.py#L375-L379.

bollwyvl:
Amazing stuff! Excited to take a look with the jupyterlab and lumino code
bases!

erikrose:
Let me know how it goes! I'm interested to get some real-world proof that it works before releasing it; there were a lot of changes!

PR #126, opened 2019-12-06
Fix some decoding crashes in Python 3.5

erikrose:


erikrose:
Ah, that's right, Python 2.7 doesn't have any of that `io` lib stuff. Never mind, at least until we drop 2.7 support, which, atm, I think is still more important than 3.5 support. If you can use 3.x, you can use a recent 3.x.

PR #137, opened 2020-04-21
Set cwd to app.confdir as expected.

bobesch:
See issue https://github.com/mozilla/sphinx-js/issues/136


erikrose:
Thank you! I'm going to fix it instead by making all paths absolute internally, since cwd is a global, brittle thing. Watch for a merge soon!

PR #141, opened 2020-05-27
Support constraints in type parameters

stasm:
We're considering sphinx-js as the documentation solution for https://github.com/projectfluent/fluent.js which is written in TypeScript. I quickly ran into #119 and here's my attempt to fix it.

I looked at the recursive approach from https://github.com/graup/sphinx-js/commit/4def189dc364f9c5a24331ae4dc7a4181abb75d5 and the other fix proposed in #124. My fix is closer to #124 except that I also went ahead and changed the return value of `make_type_name` from list to a string. It's because as far as I was able to tell, all code paths except for one return a one-element list anyways anyways.

The only case where the list can have more than one element on master is when the type is a union. I think it would actually be better to return a string for unions too, given what how other types are handled. I suspect that the current list-returning behavior is why this line was required: https://github.com/mozilla/sphinx-js/blob/404c72d01d080a228335c508c330eca0e6e6b288/sphinx_js/typedoc.py#L137

Please let me know if my assumptions are off. They're based on reading the code, the [JSDoc schema](https://github.com/jsdoc/jsdoc/blob/master/packages/jsdoc/lib/jsdoc/schema.js) and the [TypeDoc API docs](https://typedoc.org/api/interfaces/typeobject.html#name). I'm sure there's more context which I'm not aware of.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -118,45 +118,46 @@ class TypeDoc(object):
 
     def make_type_name(self, type):
         """Construct the name of a type from a Typedoc type entry"""
-        names = []
+        name = ''
         if type.get('type') == 'reference' and type.get('id'):
             node = self.nodelist[type.get('id')]
             # Should be: names = [ self.make_longname(node)]
             parent = self.nodelist[node.get('__parentId')]
             if parent.get('kindString') == 'External module':
-                names = [parent['name'][1:-1] + '.' + node['name']]
+                name = parent['name'][1:-1] + '.' + node['name']
             else:
-                names = [node['name']]
+                name = node['name']
         elif type.get('type') in ['intrinsic', 'reference', 'unknown']:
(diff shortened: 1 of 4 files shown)


stasm:
Hey @erikrose, would you have a moment to take a look at this?

erikrose:
I'm going to try to have a sphinx-js week pretty soon, and I'll look at all the open PRs then. Thanks for contributing! (I'm also looking for a maintainer for the TS parts of sphinx-js, so keep that in mind. :-))

erikrose:
I rewrote most of the TypeScript support. Along with many other issues, this one is fixed in the great IR merge of #120, demonstrated by https://github.com/mozilla/sphinx-js/blob/705659ceb08de1ec1fd16bdefd53e7b2e0a8d848/tests/test_typedoc_analysis/test_typedoc_analysis.py#L375-L379. Sorry for the delay!

PR #159, opened 2020-11-17
fix: mismatch between tsconfig.json and typedoc.json (Close #80)

### LINKED ISSUE #80: Confusion about jsdoc_config_path when using typedoc
The readme says that `jsdoc_config_path` is used for the typedoc config in typescript mode, but in fact it is used for the tsconfig.

    if app.config.jsdoc_config_path:
            jsdoc_command.add('--tsconfig', app.config.jsdoc_config_path)

I have to say, it would make more sense as explained in the readme, because you can supply tsconfig path through the typedoc config, so that way you can do both. Either way, either the readme or the code needs to be changed.

loopingz:
The parameter `jsdoc_config_path` was pointing to `typedoc --tsconfig` which is the compiler options, it should be the typedoc configuration `typedoc --options`

I added a parameter `jsdoc_tsconfig_path` to be able to specify the `typedoc --tsconfig` if needed

I updated unit test and doc.

It is a breaking change so we might want to bump version or add migration note. Available to update the PR as needed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -306,7 +306,10 @@ Configuration Reference
   A list of directories to scan (non-recursively) for JS or TS source files, relative to Sphinx's conf.py file. Can be a string instead if there is only one. If there is more than one, ``root_for_relative_js_paths`` must be specified as well. Defaults to '../'.
 
 ``jsdoc_config_path``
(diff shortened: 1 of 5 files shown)


willkg:

Issue #160, opened 2020-11-26
KeyError (ID exists) when generating documentation


azuline:
I have a problem similar to #154. Rather than the type lacking an id, however, the id seems to be missing from the `_index` property.

Traceback:
```
Traceback (most recent call last):
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/__init__.py", line 62, in analyze
    app._sphinxjs_analyzer = analyzer.from_disk(abs_source_paths,
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 40, in from_disk
    return cls(json, base_dir)
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 29, in __init__
    ir_objects = self._convert_all_nodes(json)
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 113, in _convert_all_nodes
    converted, more_todo = self._convert_node(todo.pop())
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 191, in _convert_node
    return self._convert_node(first_sig)
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 199, in _convert_node
    params=[self._make_param(p) for p in node.get('parameters', [])],
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 199, in <listcomp>
    params=[self._make_param(p) for p in node.get('parameters', [])],
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 295, in _make_param
    type=self._type_name(param.get('type', {})),
  File "/home/azul/devel/repertoire/backend/sphinx-js/sphinx_js/typedoc.py", line 240, in _type_name
    node = self._index[type['id']]
KeyError: 208
```

The typedoc output is available at https://gist.github.com/azuline/df79ece2d0b5a26958a348857d1b6b20

Issue #154, opened 2020-09-14
KeyError: 'id' processing TypeDoc


tavianator:
Just tried out sphinx-js 3.1 on https://github.com/microsoft/bistring.  I'm getting the following error:

```
Handler <function analyze at 0x7f6538653550> for event 'builder-inited' threw an exception (exception: 'id')
```

I patched sphinx up a bit to get this trace:

```
Traceback (most recent call last):
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx/events.py", line 110, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/__init__.py", line 60, in analyze
    app._sphinxjs_analyzer = analyzer.from_disk(abs_source_paths,
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/typedoc.py", line 40, in from_disk
    return cls(json, base_dir)
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/typedoc.py", line 29, in __init__
    ir_objects = self._convert_all_nodes(json)
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/typedoc.py", line 113, in _convert_all_nodes
    converted, more_todo = self._convert_node(todo.pop())
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/typedoc.py", line 160, in _convert_node
    interfaces=self._related_types(node, kind='implementedTypes'),
  File "/home/tavianator/.local/share/virtualenvs/docs-BrgkGiab/lib/python3.8/site-packages/sphinx_js/typedoc.py", line 227, in _related_types
    pathname = Pathname(make_path_segments(self._index[type['id']],
KeyError: 'id'
```

Which points [here](https://github.com/mozilla/sphinx-js/blob/8991bd2a0f8fc72d22b521fb9401e9625da717a1/sphinx_js/typedoc.py#L225).

The variable `type` looks like this:

```python
{'type': 'reference', 'typeArguments': [{'type': 'reference', 'name': 'String'}], 'name': 'Iterable'}
```

[...]

PR #162, opened 2020-12-14
typedoc: v0.19.2 adjustments

Informatic:
Notable changes:
 * "Module" has been renamed to "Namespace";
 * "External module' seems to have been renamed to "Module";
 * type id of references may be missing

All tests seem to be passing, however, as I seem to be unable to find a definitive changelog/spec of typedoc JSON output, some issues may still occur. This has already been tested on our internal codebase, where old typedoc/typescript/sphinx-js would fail. With these patches applied, support for older typedoc releases will very likely be broken. module/namespace rename could be put behind some typedoc version check.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -57,7 +57,7 @@ class Analyzer:
             if not node or node['id'] == 0:
                 # We went all the way up but didn't find a containing module.
                 return
-            elif node.get('kindString') == 'External module':
+            elif node.get('kindString') in ['External module', 'Module']:
                 # Found one!
                 return Pathname(make_path_segments(node, self._base_dir))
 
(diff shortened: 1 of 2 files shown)


Informatic:
Possibly fixes #160 / #154

Informatic:
Oops, seems like tests are failing because of typescript/typedoc upgrade, of which some transitive dependencies require Node 10, while current Travis setup is running on Node 8. All tests seemed to have passed on my Python 3.7.9 & Node v10.22.1 machine.

erikrose:
Just switched from Travis to CircleCI. Now we have node 14.15.1. :-)

Thanks for the patch! I'm not really concerned with breaking "internal module" support for old TS versions, since that flavor of module is, as I gather from people who use TS more than me, rarely used. I'm not sure how I feel about breaking old typedoc support. Do people generally keep up to date? Full review to come. Thanks again!

Informatic:
It should be possible to detect typedoc version on startup and change `Module`/`Namespace` checks to `Module`/`External module` on old enough versions.

XeniaSiskaki:
Any news on when this might get merged?

tomerghelber:
Does someone know if v0.25 and v0.19 are compatible?

Informatic:
As far as I could quickly test a couple of weeks ago, it's not and requires
more sphinx-js patches :/

willkg:

PR #169, opened 2021-02-11
Fix crashes caused by false assumptions about tsdoc output format

topiolli:
It was assumed that "sources" exists in each item of "signatures", but
this is not always the case. The fix is to look up "sources"
recursively in parent nodes.

It was also assumed that some node types have an "id", which isn't
always true.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -61,12 +61,22 @@ class Analyzer:
         node, None if one isn't found."""
         while True:
             node = node.get('__parent')
-            if not node or node['id'] == 0:
+            if not node or node.get('id', 0) == 0:
                 # We went all the way up but didn't find a containing module.
                 return
             elif node.get('kindString') == 'External module':
                 # Found one!
                 return Pathname(make_path_segments(node, self._base_dir))
+
+    def _sources(self, node):
+        """Recursively find a "sources" key in node or its parents.
+
+        Return None if no sources can be found.
+        """
+        if not node:
+            return None
+        return node.get('sources') or self._sources(node.get('__parent'))
+
 
     def _top_level_properties(self, node):
         source = node.get('sources')[0]
@@ -194,7 +204,7 @@ class Analyzer:
             # many attr of Functions.
             sigs = node.get('signatures')
             first_sig = sigs[0]  # Should always have at least one
-            first_sig['sources'] = node['sources']
+            first_sig['sources'] = self._sources(node)
             return self._convert_node(first_sig)
         elif kind in ['Call signature', 'Constructor signature']:
             # This is the real meat of a function, method, or constructor.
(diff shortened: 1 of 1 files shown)


erikrose:
It would be really helpful to have a test to demonstrate this and ensure we don't break it in the future. Do you have a small piece of code that exhibits these `sources` and `id` behaviors?

topiolli:
@erikrose: I encountered this problem while producing documentation for a relatively small TS code base, but I'm not sure if I can pinpoint the exact cause of the failure. The file that seems to be the culprit is almost 900 lines, which I don't consider "small". I can send that file if you don't mind finding out the root cause yourself. I'm not familiar with the assumptions the sphinx-js code makes about the structure, so finding out what goes wrong would be hard to me. In fact, my "fix" can very well break something else. I just changed the code so that the problems went away. :smiley: 

Now that I made changes to the aforementioned file I'm getting yet another crash. Do you have any insights on this:

~~~
Traceback (most recent call last):
  File "/home/topi/code/sphinx/sphinx/events.py", line 111, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/topi/code/sphinx-js/sphinx_js/__init__.py", line 60, in analyze
    app._sphinxjs_analyzer = analyzer.from_disk(abs_source_paths,
  File "/home/topi/code/sphinx-js/sphinx_js/typedoc.py", line 40, in from_disk
    return cls(json, base_dir)
  File "/home/topi/code/sphinx-js/sphinx_js/typedoc.py", line 33, in __init__
    self._objects_by_path.add_many((obj.path.segments, obj) for obj in ir_objects)
  File "/home/topi/code/sphinx-js/sphinx_js/suffix_tree.py", line 46, in add_many
    self.add(segs, value)
  File "/home/topi/code/sphinx-js/sphinx_js/suffix_tree.py", line 31, in add
    tree[seg] = Value(value)
TypeError: 'Value' object does not support item assignment
~~~

erikrose:
Hmm, I don't immediately see how the name of a method could make a difference. Are you saying that renaming it made the problem go away? The 'Value' object error means either our SuffixTree implementation is buggy, or our use of it is. If the former, I'm especially interested in fixing it. I'll gladly take your 900-line file! Do you want me to keep it a secret, or is it okay to attach to the ticket?

topiolli:
I realized that it is not possible to run typedoc on a single file without its dependencies and I'm not able to ship the whole code base. I however attached the typedoc output file that failed to parse.

[typedoc-output.json.txt](https://github.com/mozilla/sphinx-js/files/6147039/typedoc-output.json.txt)

erikrose:
Can you merge in from e16789fbe5a9edc6eafb96fafdf2929d2603825b and see if that gets you past the Value error? I sat down and rewrote the suffix tree, hoping that will stem the steady trickle of issues like that one.

willkg:

PR #171, opened 2021-03-16
Use app.srcdir for resolving files

letmaik:
`app.confdir` is just the folder containing the `conf.py` file and this is not necessarily the folder where the doc sources are. Incidentally using `confdir` also breaks [sphinx-multiversion](https://pypi.org/project/sphinx-multiversion/) as that specifically re-uses `conf.py` across all the doc builds by making use of sphinx' `-c` command line argument to specify the folder containing the `conf.py` file.

### DIFF
--- a/sphinx_js/__init__.py
+++ b/sphinx_js/__init__.py
@@ -44,9 +44,9 @@ def analyze(app):
     away its results in a language-specific Analyzer."""
     # Normalize config values:
     source_paths = [app.config.js_source_path] if isinstance(app.config.js_source_path, str) else app.config.js_source_path
-    abs_source_paths = [normpath(join(app.confdir, path)) for path in source_paths]
+    abs_source_paths = [normpath(join(app.srcdir, path)) for path in source_paths]
     root_for_relative_paths = root_or_fallback(
-        normpath(join(app.confdir, app.config.root_for_relative_js_paths)) if app.config.root_for_relative_js_paths else None,
+        normpath(join(app.srcdir, app.config.root_for_relative_js_paths)) if app.config.root_for_relative_js_paths else None,
         abs_source_paths)
 
     # Pick analyzer:


erikrose:
I'm open to the idea. Just to be clear, I don't assume anything in particular is in the `confdir`; I was just using it as an arbitrary base to interpret relative paths in relation to. Before we could make this change, we'd have to think out the backward-compatibility story. In all my projects, `confdir` and `srcdir` are the same, and that's what `sphinx-quickstart` seems to invariably set up as well. Is that overwhelming often the case? If so, we might be fine to just change it.

letmaik:
I don't know for sure but I haven't seen otherwise. I agree that `confdir` is as valid as `srcdir` as base to interpret relative paths, and maybe `confdir` makes more sense after all given that the sphinx-js paths (`js_source_path` etc.) are not directly related to the sphinx doc sources and it would seem more logical that relative paths in the `conf.py` file are relative to that file. Maybe the better solution then is to fix sphinx-multiversion such that it doesn't use sphinx's `-c` argument but rather copies the `conf.py` file into each branch build folder.

The hack I use at the moment is to add this to `conf.py`:
```py
def setup(self):
    self.confdir = self.srcdir
```

achamayou:
https://www.sphinx-doc.org/en/master/man/sphinx-build.html suggests that paths set by configuration values are expected to be relative to `confdir`:

> Don’t look for the conf.py in the source directory, but use the given configuration directory instead. Note that various other files and paths given by configuration values are expected to be relative to the configuration directory, so they will have to be present at this location too.

The [breathe](https://github.com/michaeljones/breathe) plugin that integrates doxygen also uses `confdir` as their base for relative paths, for what it's worth.

erikrose:
Good point: ["Note that various other files and paths given by configuration values are expected to be relative to the configuration directory"](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-c). That makes me lean toward not changing it. Breathe switched from being cwd-relative to conf.py-relative in 2014. So it would seem this is the more "normal" path.

letmaik:
Alright, makes sense, I'll take this to sphinx-multiversion.

erikrose:
Thanks! Good discussion, all!

PR #173, opened 2021-05-04
Support for JSDocs '@see {@link}' tag

xsjad0:
With the ``@see`` tag it is only possible to reference internal documented objects like functions, classes, etc.
Sphinx raises a warning if the referenced target could not be found and it just displays the non-clickable link.

```js
/**
 * @see https://github.com/mozilla/sphinx-js
 */
bad_link_function() {}
```
image

image

To reference external resources we use the ``@see {@link}`` tag a lot from within our JavaScript code. So this pull request
adds the functionality to support JSDocs ```@see {@link}``` syntax.
Here's an example:

```js
/**
 * @see {@link https://github.com/mozilla/sphinx-js}
 *
 * @returns {void}
 */
see_link_function() {}
```
image

The pull request covers the synonyms ```@linkcode``` and ```@linkplain``` though switching to a monospace font is not supported. JSDoc also provides different formats for the ```@link``` tag (see https://jsdoc.app/tags-inline-link.html). For now, I just covered the format ```@see {@link URL}```.

Notable changes are
* change docs
* implement function _prepare_see_alsos() in JSRenderer base class
* call _prepare_see_alsos() in each renderer class
* change common template
* write some test cases

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -334,8 +334,7 @@ Or, if you prefer, the Fathom repo carries a `Travis CI configuration <https://g
 Caveats
 =======
 
-* We don't understand the inline JSDoc constructs like ``{@link foo}``; you have to use Sphinx-style equivalents for now, like ``:js:func:`foo``` (or simply ``:func:`foo``` if you have set ``primary_domain = 'js'`` in conf.py.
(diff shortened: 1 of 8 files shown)


willkg:

PR #174, opened 2021-05-04
add prefix 'static' to static class methods

xsjad0:
A class method that is declared as ```static``` should differ from others in the documentation.
JSDoc adds the prefix ```static``` in front of the name of those methods. This pr adds that functionality to sphinx-js.

Here's an example of a JavaScript class containing a static and a non-static method:

```js
/**
 * SomeMembers
 */
class SomeMembers {
    /**
     * Some static member.
     *
     * @returns {void}
     */
    static someStaticMember() {
        // No use of this
    }

    /**
     * Non-static member
     *
     * @returns {void}
     */
    nonStaticMember() {
        this.nonStatic = true;
    }
}
```

The output is the following:

image

Notable changes are
* add directive staticfunction
* change template function.rst
* check if member is declared as static in analyzer class
* pass boolean variable is_static to template in renderer class
* write some tests


sphinx_js/__init__.py
```diff
@@ -31,6 +34,7 @@ def setup(app):
     app.add_config_value('js_language', 'javascript', 'env')
     app.add_config_value('js_source_path', '../', 'env')
     app.add_config_value('jsdoc_config_path', None, 'env')
+    # app.add_config_value('jsdoc_cache', None, 'env')
```
erikrose, line 37: We should wipe this out before merging.

sphinx_js/jsdoc.py
```diff
@@ -156,6 +156,10 @@ def is_private(doclet):
     return doclet.get('access') == 'private'
 
 
+def is_static(obj):
+    return obj.get('scope', '') == 'static'
```
erikrose, line 160: I understand the desire to have get() return a consistent type here, but we should either make it like is_private() (which says just `doclet.get('access')`) or change is_private() to be like `get('access', '')` for consistency.

tests/test_build_js/test_build_js.py
```diff
@@ -119,6 +119,23 @@ def test_autofunction_see(self):
             '     * "deprecatedFunction"\n\n'
             '     * "DeprecatedAttribute"\n')
 
+    def test_staticfunction(self):
```
erikrose, line 122: Does `js:staticfunction` need to be exposed to the user? I hope it might be an implementation detail and users could simply use `static` in their JS code. Perhaps there are corner cases where people would monkeypatch static methods onto objects, making it impossible to declare them `static`, but I would hope those are rare.

If it is an implementation detail, we can dispense with this test and staticfunction.rst.

erikrose (changes requested):
A nice, clean addition! I just want to understand whether `staticfunction` needs to be user facing. (And if it does, we need to document it in the readme.) Thanks!

sphinx_js/__init__.py
xsjad0: You're right! I removed it in a new commit.
Maybe i committed that line in other prs too -.-

sphinx_js/jsdoc.py
xsjad0: Think so too. For consistency, I changed it to be like the is_private method. Basically, there's no change for us if we compare None or '' with 'static' 👍. I also changed the parm ``obj`` to ``doclet``, so we're even more consistent.

tests/test_build_js/test_build_js.py
xsjad0: I'm not quite sure.. But let's say someone really wants to define a static method in a .rst file by hand (not based on JS code), we should propably expose the new directive.. 

By the way, the python domain also exposes a ``py:staticmethod`` directive. It is also possible to use ``py:method`` with the option ``:staticmethod:``  (see [sphinx-docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=directive#directive-py-method)). To be more consistent with sphinx in general, maybe it's a good idea.

erikrose: > But let's say someone really wants to define a static method in a .rst file by hand (not based on JS code),

That's a good point.

Taking a look at "the way the wind is blowing" in Sphinx, I see `py:staticmethod` was introduced in 0.4 and the `:staticmethod:` option in 2.1. Corresponding top-level directives have not been introduced since then for similar options like `:final:`, `:abstractmethod:`, and `:async:`. I think the wind is blowing in the direction of options, and we should go with it. Also, options allow you to combine prefixes, like `async static final`.

The ideal, then, would be to support just a `:static:` option on `js:function`. Of course, we don't own `js:function`. Here's what I'd recommend:

1. For now, let's merge your patch but not tell anyone about `js:staticmethod`. Let it be a private directive which we'll ultimately aim to obsolete. And let's delete the test that makes sure it's there; the other one will act as a sufficient integration test. (If you can figure out how to add the prefix without having a separate directive, even better.)
2. In Sphinx itself, add a `:staticmethod:` option to `js:method`, where it belongs. Once that merges in, we can start using it.

Looking farther ahead, I think sphinx-js will eventually stop using `js:function` and instead provide its own rendering. That will allow us to do things not dreamt of by Sphinx, like hyperlinking TypeScript type names to their definitions.

What do you think?

xsjad0: Options are definitely the better approach! But as you said, we don't own the ``js:function`` directive.
Taking a look at the JS domain source code in sphinx, here [define prefix](https://github.com/sphinx-doc/sphinx/blob/512743525c64ef880ef02116fe3890ed29339bd4/sphinx/domains/javascript.py#L230) and [add prefix](https://github.com/sphinx-doc/sphinx/blob/512743525c64ef880ef02116fe3890ed29339bd4/sphinx/domains/javascript.py#L94), the sphinx-team did the same to get JS classes to work. 
Unfortunatly, I can't see any chance to hook in and add the prefix at runtime when we would need it.
Looking at the python implementation of [py:method](https://github.com/sphinx-doc/sphinx/blob/512743525c64ef880ef02116fe3890ed29339bd4/sphinx/domains/python.py#L708), they checked the directives options to add any combination of supported prefixes. 

So, I'm completely with you. In Future, we should be able to use options. For now, I'll drop that test and we're able to merge. 

> In Sphinx itself, add a :staticmethod: option to js:method, where it belongs. Once that merges in, we can start using it.

That might be a small effort, I'll try to do this soon 👍

erikrose:
Thanks so much, Marius! It'll be in the next release.

PR #175, opened 2021-05-04
add js autodoc default options

xsjad0:
This allows us to omit the options ``:members:`` and ``private-members`` on every single autoclass directive.

conf.py
```python
js_autodoc_default_options = {
    "members": True,            # by default False
    "private-members": True,    # by default False
}
```

JavaScript source
```js
/**
 * TestClass
 */
class TestClass {
    /**
     * A.
     */
    a() {}

    /**
     * B.
     */
    b() {}

    /**
     * C.
     * @private
     */
    c() {}
}
```

.rst file
```code
.. js:autoclass:: TestClass
```

html result

image

Notable changes:
* add config value js_autodoc_default_options (default empty dict)
* implement _default_options() in JSRenderer class
* try to set defaults on initialization of renderer classes (currently only at AutoClassRenderer)
* write some test cases

### DIFF
--- a/tests/test_build_js/source/code.js
+++ b/tests/test_build_js/source/code.js
@@ -1,4 +1,3 @@
-/* eslint-disable */
 /**
  * Return the ratio of the inline text length of the links in an element to
  * the inline text length of the entire element.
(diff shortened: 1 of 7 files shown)


tests/test_build_js/source/code.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
xsjad0, line 1: I'm going to wipe this out -.-

willkg:

PR #176, opened 2021-05-05
Add support for JSDocs @namespace tag

xsjad0:
Instead of using classes to encapsulate some JS code we often use namespaces.
To document namespaces JSDoc offers the ``@namespace`` tag. 
The documentation looks quite similar to a class documentation.
At first glance, there's just another prefix (namespace instead of class). 

Example JS code

```js
/**
 * A simple namespace.
 * @namespace
 */
var NewNamespace = {
    /** documented as NewNamespace.someAttribute. */
    someAttribute : "foo",

    /**
     * documented as NewNamespace.foo
     * @returns {void}
     */
    foo : function() {},

    /**
     * @returns {void}
     */
    bar : function() {},

    /**
     * @private
     * @returns {void}
     */
    baz : function() {},
};
```

Document namespace like this
```code
.. js:autonamespace:: NewNamespace
    :members:
    :private-members:
    :exclude-members: bar
```
Generated output

image

Notable changes are:
* add template namespace.rst
  * similar to class.rst
  * no supers, no interface, no is_interface, no is_static
* add directives ``namespace`` and ``autonamespace`` to js domain
  * autonamespace and autoclass have the same options
* add ir class Namespace
* collect doclets by namespace
* add renderer AutoNamespaceRenderer
* write some tests (very similar to autoclass tests)

I'm not sure, if this directive is going to be used by a lot of people. We just wanted to see a slightly different documentation for classes and namespaces. In addition, the changes do no harm on the rest of the code^^

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -260,7 +260,7 @@ Finally, if you want full control, pull your namespace members in one at a time 
 
 To reference a namespace from anywhere in your documentation, you can use the ``js:ns`` role
 
-    :js:ns:`SomeNamespace`
+    ``:js:ns:`SomeNamespace```
 
 autoattribute
 -------------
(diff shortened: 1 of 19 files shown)


willkg:

PR #178, opened 2021-05-12
bugfix @see tag does not work with static function/member

xsjad0:
Using JSDocs ``@see`` tag to reference a static function/member causes Sphinx to crash with the following error message

image

Since we've added the directive ``js:staticfunction`` we also have to specify a new object type within the JS Domain that allows Sphinx to resolve the any reference.

This example code works now:
```js
    /**
     * SomeMembers
     */
    class SomeMembers {
        /**
        * Some static member.
        *
        * @returns {void}
        */
        static someStaticMember() {
            // No use of this
        }

        /**
        * Non-static member
        *
        * @see someStaticMember
        * @returns {void}
        */
        nonStaticMember() {
            this.nonStatic = true;
        }
    }
```

It's also possible to reference a static function/member by using the ``js:func:`` role 

```code
    :js:func:`SomeMembers.someStaticMember`
```

Unfortunately, there is no way extend Sphinx to define the correct index text for static functions.

See sphinx.domains.javascript.py code snippet
```python
def get_index_text(self, objectname: str, name_obj: Tuple[str, str]) -> str:
    name, obj = name_obj
    if self.objtype == 'function':
        if not obj:
            return _('%s() (built-in function)') % name
        return _('%s() (%s method)') % (name, obj)
    elif self.objtype == 'class':
        return _('%s() (class)') % name
    elif self.objtype == 'data':
        return _('%s (global variable or constant)') % name
    elif self.objtype == 'attribute':
        return _('%s (%s attribute)') % (name, obj)
    return ''
```

Unless we generate a JS index, we may not notice any restriction...
What do you think??

### DIFF
--- a/sphinx_js/__init__.py
+++ b/sphinx_js/__init__.py
@@ -1,6 +1,9 @@
 from os.path import join, normpath
 
+from sphinx.domains.javascript import JavaScriptDomain
+from sphinx.domains import ObjType
 from sphinx.errors import SphinxError
+from sphinx.locale import _
 
 from .directives import (auto_class_directive_bound_to_app,
                          auto_function_directive_bound_to_app,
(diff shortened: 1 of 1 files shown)


willkg:
I redid the suggested fix in #217 after going through and making sure there weren't any changes in Sphinx that would cause us to do something else. Thank you for the work sleuthing it out!

PR #180, opened 2021-09-08
Fixed errors on convering typedoc JSON output to IR object

ruslankl9:
This is bugfix of converting typedoc json output.
I found that `KeyError: 'id'` raising with 'extendedTypes' objects, not having 'id' key. This happens on processing code with extending classes like:
```
import { SomeClass } from 'module';

class MyClass extends SomeClass {
  //...
}
```
Also `KeyError: 'sources'` raising with Constructor kind when class definition doesn't have one. Example:
```
class SomeClass {}
```
In this case we are getting 'sources' value from the parent Class kind object.


willkg:

Issue #186, opened 2022-01-05
incompatible with python 3.10


2bndy5:
This is the third time I've run into this error using python 3.10.
```
Could not import extension sphinx_js (exception: cannot import name 'Mapping' from 'collections' (C:\Users\%user_name%\AppData\Local\Programs\Python\Python310\lib\collections\__init__.py))
```

Apparently, `Mapping` no longer exists (as an alias) in the `collections` module.

The backward compatible fix that I've been suggesting to other python devs hitting the same problem is
```py
try:
    from collections.abc import Mapping
except ImportError:
    # TODO: remove when dropping python 2.7
    from collections import Mapping
```
 However, I can't narrow down where exactly this would be applied. When doing a search for the keyword collections, I only come up with 

https://github.com/mozilla/sphinx-js/blob/a1ee45651f14bba63a4e2e370e9acadc847c017b/sphinx_js/jsdoc.py#L8

and also the same line in a file called doclets.py (I can't find that file here on github though).

I think `defaultdict` implicitly looks for Mapping from collections, so I have no idea about how to solve this without taking this upstream (trying to figure out [python's bug tracker](https://bugs.python.org/) seems like too much hassle).

2bndy5:
Found this comment in the python issue tracker:
> According to https://docs.python.org/3.9/library/collections.html, aliases like collections.Mapping were removed from the collections namespace in 3.10.

2bndy5:
I think this is a [sphinx-js dependency](https://github.com/mozilla/sphinx-js/blob/a1ee45651f14bba63a4e2e370e9acadc847c017b/setup.py#L17) problem (concerning parsimonious pkg). 

<details><summary>Full trace back from <code>sphinx-build . _build</code></summary>

```
Original exception:
Traceback (most recent call last):
  File "C:\Users\%user_name%\Documents\GitHub\venv\lib\site-packages\sphinx\registry.py", line 429, in load_extension
    mod = import_module(extname)
  File "C:\Users\%user_name%\AppData\Local\Programs\Python\Python310\lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
[...]

PR #184, opened 2021-11-29
Allow py39+ compatible `parsimonious` version

webknjaz:
Currently, the users of sphinx-js will only be able to get parsimonious v0.7.0 which is about
5 years old and does not support Python 3.9 or higher. The newer versions of the 0.8.x
stream fix the compatibility issues and are necessary `sphinx-js` to be usable under
Python 3.10 and Python 3.9.

PR #204, opened 2022-11-04
Update dependencies for python 3.10 compat and fix an exception

dhdaines:
Sorry, two changes in one...

- parsimonious 0.7.0 doesn't work with Python 3.10 (fixes https://github.com/mozilla/sphinx-js/issues/186)
- trying to report an exception in `JsRenderer.dependencies` causes an exception because the exception isn't the exception we're looking for

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     packages=find_packages(exclude=['ez_setup']),
     url='https://github.com/mozilla/sphinx-js',
     include_package_data=True,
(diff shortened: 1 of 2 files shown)


2bndy5:
FYI, it looks like #184 also has a fix for #186. IMHO it is more backwards compatible than the fix here.

lonnen:
I've hoisted the JsRenderer commit out and landed it in: https://github.com/mozilla/sphinx-js/pull/206

The parsimonious change has been landed a different way. Note: this is still probably not going to work right in Python 3.10 yet but its getting closer

lonnen:
Thank you for the contribution!

PR #207, opened 2022-11-28
run multiple versions of python 3.x in CI and Tox

lonnen:
Tox has only been wired up to run 3.9 in CI

This expands CI coverage to supported version of Python

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,9 +8,11 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  test:
 
     runs-on: ubuntu-latest
+
+    continue-on-error: true
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


lonnen:
It's possible to mark support for some versions as "experimental" so that the test outcomes don't contribute to the overall status of the build. It's not immediately obvious to me if this is useful for us. We might consider 3.11 experimental, but 3.10+ is failing and we really want to get to 3.10

lonnen:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs describes a configuring using "experimental" but it's less a built in feature of the platform and more a generic feature with specific documentation. Once implemented it basically ends up where it was in the UI and from a downstream consumers perspective. It may be clearer in the actual configuration of the CI. Unclear if this is worth the complexity trade. Sleeping on it

lonnen:
merging this in without 3.10+ versions in order to expand coverage as much as we can immediately

PR #208, opened 2022-11-29
restore python versions 3.10 and 3.11 in the test matrix

lonnen:
This reverts commit 17e2ca156a9461dc680616ac0915f6e1ebbb231d and restores 3.10 and 3.11 to the test matrix. At the time this is opened, these are expected to fail and should not be merged until they are passing.

For more context, see: https://github.com/mozilla/sphinx-js/pull/207


lonnen:
this was done in #210

Issue #203, opened 2022-09-23
ImportError: cannot import name 'soft_unicode' from 'markupsafe'


diogobaltazar:
# Issue

Can't build the documentation with sphinx when installing the module ``sphinx-js==3.1.2``.

# Reproduce

```sh
(.env) python
>>> import sphinx_js
...
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (.env/lib/python3.8/site-packages/markupsafe/__init__.py)
```

# Hints

Version ``sphinx-js==3.1.2`` has a dependency on ``markupsafe==2.0.1`` which has removed ``soft_unicode`` (see [Version 2.1.0](https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0) release notes).

# Resolved

Resolved by reverting the version to ``2.0.1`` with ``pip install markupsafe==2.0.1``

willkg:
This was fixed in abb38d2e.

PR #210, opened 2022-12-10
fixes sphinx-js with sphinx 4.1.0+ (#209, #203)

### LINKED ISSUE #209: sphinx-js incompatible with Sphinx 4.1.0+
Sphinx-js cannot simply update to use the latest version of sphinx-doc. This is preventing sphinx-js from supporting python 3.10

When we use that commit or later versions of Sphinx the sphinx-js test suite incorrectly prints argument, error, and return types as `"type()"` instead of the expected `*type*`. It looks something like this:

```
tests/test_build_js/test_build_js.py:50: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <test_build_js.Tests testMethod=test_autofunction_callback>
filename = 'autofunction_callback'
contents = 'requestCallback(responseCode)\n\n   Some global callback\n\n   Arguments:\n      * **responseCode** (*number*) --\n'

    def _file_contents_eq(self, filename, contents):
>       assert self._file_contents(filename) == contents
E       AssertionError: assert ('requestCallback(responseCode)\n'\n '\n'\n '   Some global callback\n'\n '\n'\n '   Arguments:\n'\n '      * **responseCode** ("number()") --\n') == ('requestCallback(responseCode)\n'\n '\n'\n '   Some global callback\n'\n '\n'\n '   Arguments:\n'\n '      * **responseCode** (*number*) --\n')
E           requestCallback(responseCode)
E           
E              Some global callback
E           
E              Arguments:
E         -       * **responseCode** (*number*) --
E         ?                           ^      ^
E         +       * **responseCode** ("number()") --
E         ?                           ^      ^ ++
```

After [some investigation](https://github.com/mozilla/sphinx-js/pull/191), I've narrowed the incompatibility to Sphinx-doc 4.1.0 and newer, specifically https://github.com/sphinx-doc/sphinx/commit/4534d2d1a5755c8cbc9ef4327eab7e34a85a7de8. This was shipped as part of [sphinx-doc PR 9155](https://github.com/sphinx-doc/sphinx/pull/9155), which resolved [Sphinx-doc issue 8945](https://github.com/sphinx-doc/sphinx/issues/8945#issuecomment-812981784).

The commit contains a change to the [docfields.py:make_xref function](https://github.com/sphinx-doc/sphinx/commit/4534d2d1a5755c8cbc9ef4327eab7e34a85a7de8#diff-f18bf3af0871669d46455026b741254522ecd7e8a729be708603249e872d37abR66-R94). If I revert just this portion of the change things work as expected.

In the previous code the role is present but ignored. The prior code path is preserved there but there are no conditions under which it can be reached. The inliner is always passed along with domain and role ('js:function'). The test failures we see are all `xref` objects that resolve through the new code as functions, which is why they get inappropriately rendered with parenthesis.

It's not obvious to me what should or even can be done about this from the sphinx-js side of the code.

willkg:
I had to fix some minor things before I was able to work on this, but the bulk of the PR is a workaround for [sphinx issue #11021](https://github.com/sphinx-doc/sphinx/issues/11021).

The gist of it is that this fixes the javascript domain `TypedField` and `GroupedField` to do what the python domain equivalents do--namely set `inliner` to `None` before calling `Field.make_xref`.

@lonnen's work finding the problematic commit in Sphinx was invaluable towards this fix. That made this so much easier than when I looked at it a year ago and bounced off it.

This also pulls in @lonnen's fix from PR #191 because that fixed a couple of tests that weren't working with whatever version of Sphinx I was testing with that was > 4.1.0.

The end result is that sphinx-js tests pass in Python 3.7 through 3.11 with Sphinx 4.10 through 5.3.0.

I thought about different ways to fix the problem. It's tricky because the bit we need to fix isn't very accessible. I tried to make this as minimal, but maintainable, as possible. If the comments are opaque, we should improve them. Future us will want present us to do that.

Pretty sure this fixes #209 and #203. It also replaces #191 and #208.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,11 @@ setup(
     packages=find_packages(exclude=['ez_setup']),
     url='https://github.com/mozilla/sphinx-js',
     include_package_data=True,
-    install_requires=['Jinja2>2.0,<3.0', 'markupsafe==2.0.1', 'parsimonious>=0.9.0,<0.10.0', 'Sphinx>=3.0.0'],
+    install_requires=[
+        'Jinja2>2.0',
+        'parsimonious>=0.10.0,<0.11.0',
+        'Sphinx>=4.1.0,<6.0.0'
+    ],
     python_requires='>=3.7',
     classifiers=[
         'Intended Audience :: Developers',
@@ -24,6 +28,6 @@ setup(
         'Programming Language :: Python :: 3',
         'Topic :: Documentation :: Sphinx',
         'Topic :: Software Development :: Documentation'
-        ],
+    ],
     keywords=['sphinx', 'documentation', 'docs', 'javascript', 'js', 'jsdoc', 'restructured', 'typescript', 'typedoc'],
 )
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,9 @@ setup(
         'Jinja2>2.0',
         'parsimonious>=0.10.0,<0.11.0',
         'Sphinx>=4.1.0,<6.0.0'
+        # Pin markupsafe because of
+        # https://github.com/pallets/jinja/issues/1585
+        'markupsafe==2.0.1',
     ],
     python_requires='>=3.7',
     classifiers=[
--- a/sphinx_js/__init__.py
+++ b/sphinx_js/__init__.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from os.path import join, normpath
 
 from sphinx.errors import SphinxError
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -14,7 +14,11 @@
     packages=find_packages(exclude=['ez_setup']),
     url='https://github.com/mozilla/sphinx-js',
     include_package_data=True,
-    install_requires=['Jinja2>2.0,<3.0', 'markupsafe==2.0.1', 'parsimonious>=0.9.0,<0.10.0', 'Sphinx>=3.0.0'],
```
willkg, line 17: markupsafe isn't directly used by sphinx-js--pretty sure it's a secondary dependency--so I removed it.

setup.py
```diff
@@ -14,7 +14,11 @@
...
+    install_requires=[
+        'Jinja2>2.0',
+        'parsimonious>=0.10.0,<0.11.0',
+        'Sphinx>=4.1.0,<6.0.0'
```
willkg, line 20: I picked an upper bound for Sphinx because they're expecting 6.0.0 to drop support for Python 3.7 and I don't know the future, but figured maybe this saves someone some time.

I also updated parsimonious version requirement here to the one that supports Python 3.10 and 3.11.

sphinx_js/__init__.py
```diff
@@ -9,6 +10,68 @@
 from .jsdoc import Analyzer as JsAnalyzer
 from .typedoc import Analyzer as TsAnalyzer
 
+
+# Cache this to guarantee it only runs once.
+@lru_cache(maxsize=None)
```
willkg, line 15: I figured I'd guard against this running multiple times even though "normal use" probably won't run this multiple times. I added the comment because this is a frankly bizarre use of `lru_cache`.

sphinx_js/__init__.py
```diff
@@ -9,6 +10,68 @@
...
+    class JSXrefMixin:
+        def make_xref(
+            self,
+            rolename,
+            domain,
+            target,
+            innernode=nodes.emphasis,
+            contnode=None,
+            env=None,
+            inliner=None,
```
willkg, line 36: Because this defines inliner and location, if we want this fix to work with Sphinx < 4.1.0, we're going to have to detect the Sphinx version and then have a different mixin for < 4.1.0.

Do we need to support Sphinx < 4.1.0? Should we do this?

tox.ini
```diff
@@ -13,9 +13,12 @@ python =
 setenv =
     PATH={toxinidir}/node_modules/.bin{:}{envbindir}{:}{env:PATH}
 deps = -rrequirements_dev.txt
+allowlist_externals =
```
willkg, line 16: `allowlist_externals` is what replaced `whitelist_externals` for recent versions of tox.

setup.py
lonnen: Markupsafe is a secondary dependency. It did not follow semver and became incompatible with Jinja at one point.
https://github.com/pallets/jinja/issues/1585

The recommended path forward has been upgrading rather than pinning, but I approved this pin in order to get the tree back to green with the intention to fast follow on getting Jinja, Sphinx, and Parsimonious updated. Good to remove it here while you're in there updating things

sphinx_js/__init__.py
lonnen: I'm not aware of a reason to support older versions of Sphinx

lonnen (approved):
Thank you Will! I've left some notes in case I need to revisit this in the future but it looks good to land as-is. 

Long term I hope we see a fix in sphinx. Thank you for your notes over [there](https://github.com/sphinx-doc/sphinx/issues/11021). I suspect that will make it much more likely that it gets fixed and we can remove this later.

setup.py
willkg: Got it. I will re-add it with a comment because we don't know what's going to happen going forward.

willkg:
Merged! Thank you for the review!

PR #211, opened 2022-12-10
Clarify usage in README (#185)

### LINKED ISSUE #185: Some clarification needed in README
"write a folder full of reStructuredText files"
Is a slightly confusing command unless you are already familiar with sphinx and do not use apidoc.

An anyone give a working example of these files.

willkg:
Fixes #185.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,7 @@ Setup
 Use
 ===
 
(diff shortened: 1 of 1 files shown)


willkg:
self-r+. It's a minor docs change.

PR #212, opened 2022-12-10
Fix Sphinx dependency

willkg:
The missing comma causes the two strings to concatenate. This was a bug from PR #210. It's weird that it didn't get picked up in CI.


willkg:
self-r+

PR #216, opened 2022-12-16
Fix tests to support jsdoc 4.0.0 (#215)

### LINKED ISSUE #215: support jsdoc 4.0.0
Latest version of jsdoc is 4.0.0 and it was released in November 2022. sphinx-js tests fail when we update tests to test against jsdoc 4.0.0 output.

This covers fixing that.

willkg:
Seems like everything is ok. Some of the test assertions changed values slightly, so I added support for asserting both values even though we're not testing with both versions of jsdoc. We can drop the old stuff later.

Fixes #215.

### DIFF
--- a/tests/test_jsdoc_analysis/test_jsdoc.py
+++ b/tests/test_jsdoc_analysis/test_jsdoc.py
@@ -86,8 +86,16 @@ class ClassTests(JsDocTestCase):
         assert cls.path == Pathname(['./', 'class.', 'Foo'])
         assert cls.filename == 'class.js'
         assert cls.description == 'This is a long description that should not be unwrapped. Once day, I was\nwalking down the street, and a large, green, polka-dotted grand piano fell\nfrom the 23rd floor of an apartment building.'
-        assert cls.line == 15  # Not ideal, as it refers to the constructor, but we'll allow it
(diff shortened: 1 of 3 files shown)


tests/test_jsdoc_analysis/test_jsdoc.py
```diff
@@ -86,8 +86,16 @@ def test_class(self):
         assert cls.path == Pathname(['./', 'class.', 'Foo'])
         assert cls.filename == 'class.js'
         assert cls.description == 'This is a long description that should not be unwrapped. Once day, I was\nwalking down the street, and a large, green, polka-dotted grand piano fell\nfrom the 23rd floor of an apartment building.'
-        assert cls.line == 15  # Not ideal, as it refers to the constructor, but we'll allow it
-        assert cls.examples == ['Example in constructor']  # We ignore examples and other fields from the class doclet so far. This could change someday.
+        # Not ideal, as it refers to the constructor, but we'll allow it
+        assert cls.line in (
+            8,  # jsdoc 4.0.0
```
willkg, line 91: The changes in the test values suggests that maybe this fixes the issue in the comment where with jsdoc 3.6.3, we were getting data about the constructor and now in jsdoc 4.0.0, we're getting data about the class.

willkg:
self-r+

PR #217, opened 2022-12-16
Fixes objtype for staticfunction (#178)

willkg:
This fixes the objtype xref when the thing is a staticfunction that was added to sphinx-js in pr #174.

Now `@see` should work and the index should show the right text.

This is based on work from @xsjad0 and obsoletes #178.

### DIFF
--- a/tests/test_build_js/source/code.js
+++ b/tests/test_build_js/source/code.js
@@ -242,6 +242,15 @@ class SimpleClass {
 
     /**
      * Static.
+     *
+     * @see nonStaticMethod
      */
-    static noUseOfThis() {}
+    static staticMethod() {}
+
+    /**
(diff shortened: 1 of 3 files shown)


tests/test_build_js/source/code.js
```diff
@@ -242,6 +242,15 @@ class SimpleClass {
...
+     * @see nonStaticMethod
+     */
+    static staticMethod() {}
+
+    /**
+     * Non-static member.
+     *
+     * @see staticMethod
      */
-    static noUseOfThis() {}
```
willkg, line 246: I changed the name so the tests are easier to read. `noUseOfThis` suggests to me I shouldn't be seeing it in test output, but it's intentionally there.

PR #219, opened 2023-01-12
Fix support for typedoc 0.19

hoodmane:
Before this PR:
typedoc v0.15 and 0.16 work
>= 0.17 crash (cannot render anything due to `ValueError('Could not find deppath')`)

After this PR:
v0.15 and 0.16 still work
v0.17 and 0.18 mess up the type name for tuple types due to a bug in these versions of typedoc
TypeNameTests::test_basic - AssertionError: assert '<TODO: other type>' == '[string, number]'
v0.19 works
v0.20+: crash (cannot render anything because "sources" key not always present on nodes and there is another deppath problem).

@rth @ryanking13

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -64,7 +64,7 @@ class Analyzer:
             if not node or node['id'] == 0:
                 # We went all the way up but didn't find a containing module.
                 return
-            elif node.get('kindString') == 'External module':
+            elif node.get('kindString') in ('Module', 'External module'):
                 # Found one!
                 yield node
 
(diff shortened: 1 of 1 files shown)


hoodmane:
It would be nice to add a github workflow matrix that tests against different versions of typedoc and a fixed version of Python.

hoodmane:
I think the correct definition for `_containing_deppath` would be simply:
```py
 def _containing_deppath(self, node) -> str:
     return node["sources"][0]['fileName']
```
The hack I have in the current PR stops working in v0.20

willkg:

PR #220, opened 2023-01-30
Update sphinx-quickstart tool output in README.md

Firfi:
The README.md output is for an older version and they changed seemingly a lot in https://github.com/sphinx-doc/sphinx/issues/4148 . 

For instance, setting it up, I was a bit confused thinking I could select the documentation directory in the cli tool. Now instead, it conveniently auto-selects the root project directory `.` to put its files in.

This PR clarifies this a bit to match expectations for the tool output with the latest version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,29 +33,28 @@ Setup
         cd my-project
         sphinx-quickstart
 
-          > Root path for the documentation [.]: docs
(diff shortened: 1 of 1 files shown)


willkg (approved):
Thank you!

Issue #233, opened 2023-09-28
transfer sphinx-js project


willkg:
Issue #226 caused us to sit down and figure out the future of this project. @hoodmane is maintaining a fork for pyodide. We worked out that we're going to transfer this project to @hoodmane. 

This covers the tasks to transfer:

* [x] remove `CODE_OF_CONDUCT.md`
* [x] go through and remove Mozilla branding--this will no longer be a Mozilla-owned project
* [x] remove author names where appropriate with a TODO to fill new ones in
* [ ] transfer project to new owner
* [ ] update permissions
* [ ] update pypi permissions

hoodmane:
> remove CODE_OF_CONDUCT.md

Seems like a fine code of conduct but I guess it is very Mozilla branded. I guess we can copy our own version from here:
https://github.com/pyodide/pyodide/blob/main/CODE-OF-CONDUCT.md

>  remove author names where appropriate

I'm always confused about what the etiquette is with this for forks and ownership transfers. I think that the original authors are still authors but no longer maintainers. The primary argument for removing old people is if they want to be removed for some reason or if they are getting contacted by people about it? What do you think?

cc @rth @ryanking13 @erikrose

Thanks again @willkg for your work on this project!

willkg:
In other projects, I have a contributors file where people who've contributed are listed. This project doesn't do that, but it could. We could seed it with the git committers.

I find "authors" and "maintainers" project metadata which gets tracked in various places to be an invitation for users to contact when they need help. Thus, people who aren't active on a project shouldn't be listed there.

I'm not aware of any virtues to being listed in "authors" or "maintainers" but not being active.

hoodmane:
The only reason I'm sensitive of this point is that I observed an argument about this point in the past: a new maintainer removed the previous maintainer from the authors list and the old maintainer got angry. 

> "authors" and "maintainers" project metadata which gets tracked in various places to be an invitation for users to contact when they need help. Thus, people who aren't active on a project shouldn't be listed there.

This seems like a good argument.

willkg:
I respect the work Erik put into sphinx-js and you're right--we shouldn't just drop things like that entirely. I added a CONTRIBUTORS file like [I have with Bleach](https://github.com/mozilla/bleach/blob/main/CONTRIBUTORS) in PR #234. Does that look ok?

hoodmane:
Looks great.

rth:
[...]

PR #234, opened 2023-09-28
Remove Mozilla-centric things (#233)

willkg:


willkg:
@lonnen I think this covers everything we should do before transferring. Can you think of anything else?

CONTRIBUTORS
```diff
@@ -0,0 +1,36 @@
+Bleach was originally written and maintained by Erik Rose and various
```
hoodmane, line 1:
```suggestion
sphinx-js was originally written and maintained by Erik Rose and various
```

CONTRIBUTORS
```diff
@@ -0,0 +1,36 @@
...
+contributors within and without the Mozilla Corporation and Foundation.
+
+It is currently maintained by XYZ.
```
hoodmane, line 4:
```suggestion
It is currently maintained by Hood Chatham.
```
Or I can insert this myself later. Thanks!

CONTRIBUTORS
willkg: Thank you!

CONTRIBUTORS
willkg: Done!

lonnen (commented):
If we're removing `CODE_OF_CONDUCT.md` we should also remove it from `MANIFEST.in`

lonnen:
oops, this was already merged. I'm on it

hoodmane:
I'm planning to add the code of conduct from Pyodide so it's probably not important. But also do we need to put the code of conduct into the wheel?

PR #236, opened 2023-10-10
fixup follow up to PR #234

lonnen:
remove the alread-removed Code_of_Conduct line from the manifest

My bad for missing the review request earlier. This covers it


willkg (approved):
Good catch!

PR #255, opened 2024-12-24
Fix flake8 so it runs and then fix linting issues

willkg:


cclauss:
Perhaps upgrade to ruff... https://docs.astral.sh/ruff/linter

willkg:
This project is about to be deprecated. I won't be doing any further development on it. Right now I'm trying to get things working so I can cut the last release.

PR #280, opened 2025-09-05
Fix running tests with node 24

Standard8:
This picks up the latest version of jsdoc which fixes the compatibility issue.

Without this, there are many errors along the lines of:

```
/Users/mark/dev/sphinx-js/.nox/tests-3-13/node_modules/jsdoc/lib/jsdoc/util/dumper.js:98
        else if ( util.isRegExp(o) ) {
                       ^

TypeError: util.isRegExp is not a function
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 24
 
       - name: Set up Python
         uses: actions/setup-python@v4.3.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v7
         with:
-          node-version: 22
+          node-version: 24
 
       - name: Set up Python
         uses: actions/setup-python@v7.0.0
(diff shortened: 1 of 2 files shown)


hoodmane:
Is it also a problem in node 24? Can we run node 24 in ci and see the problem?

Standard8:
@hoodmane I've just bumped the node version for tests to v24. I noticed I was also running 24.5.0 locally - I referenced 23 in the commit message, as that's what JSDoc reference when they fixed it.

hoodmane (approved):

hoodmane:
pre-commit failure doesn't seem related to this PR...

hoodmane:
But the other CI jobs need to be fixed.

Standard8:
@hoodmane I've updated the PR to the latest `master` so that the test logs can be updated as they had expired. Would you be able to trigger them again, so that I can investigate the issue?

hoodmane:
Done.

Standard8:
Thank you, that was quick.

I found the cause - I'd missed updating the `jsdoc` version in a couple of places. That is now fixed. I don't know if we want to continue with node 24 or bump to 26 or latest or something. Happy to update if you want to.

hoodmane:
Start with merging node 24 checks and then we can do node 26?

Standard8:
Sure

hoodmane:
Thanks @Standard8!

PR #284, opened 2025-09-17
FIX Fix a bug that comment of the arrow function in the interface is not rendered

ryanking13:
This fixes a bug while I was working on https://github.com/pyodide/pyodide/pull/5746.

I noticed that a docstring of a property inside an interface that has arrow function is not rendered in the docs.

For example,

```typescript
interface MyInterface {
  /**
    * This should be rendered
    */
  prop: () => void;
}
```

The docstring "This should be rendered" is not rendered in the docs.


sphinx_js/js/convertTopLevel.ts
```diff
@@ -537,7 +537,18 @@ export class Converter {
...
       // TODO: unclear if this is the right behavior. Maybe there should be a
       // way to pick?
-      return [this.functionToIR(prop.type.declaration), []];
+      const functionIR = this.functionToIR(prop.type.declaration);
+
+      // Preserve the property's own documentation if it exists
+      const propertyDescription = renderCommentSummary(prop.comment);
+      if (propertyDescription.length > 0) {
+        functionIR.description = propertyDescription;
+      }
```
hoodmane: What about:
```suggestion
      functionIR.description = renderCommentSummary(prop.comment);
```

hoodmane (approved):
Thanks!

sphinx_js/js/convertTopLevel.ts
ryanking13: Yeah, that looks simpler

ryanking13:
Thanks for the review!

PR #287, opened 2025-10-16
Unpin markupsafe

fmhoeger:
Currently, markupsafe is pinned to version 3.0.2.

This PR removes the `markupsave` version constraint.


hoodmane:
Thanks! Do we know we need < 4? What happens if we just remove the cap?

fmhoeger:
Yes, we could also just remove the restriction completely as version 4 doesn't even exist yet.

hoodmane (approved):

PR #297, opened 2025-12-28
Include node on typedoc command

### LINKED ISSUE #285: Windows build fails with "WinError 193" when js_language = 'typescript'
Building documentation with `sphinx-js` on Windows fails with an `OSError: [WinError 193] %1 is not a valid Win32 application` error when `js_language = 'typescript'` is set in `conf.py`. The build proceeds without error when `js_language` is left at its default or explicitly set to `'javascript'`. This indicates an issue with how `sphinx-js` invokes the Node.js subprocess with TypeDoc on Windows.

### Steps to Reproduce

1.  **Environment:**
    *   **OS:** Windows 10/11 (x64)
    *   **Python:** 3.x (x64)
    *   **Node.js:** 18+ (x64)
    *   **`sphinx-js`:** Latest version from `pip`
    *   **`typedoc`:** Latest version from `npm`
2.  **`conf.py` setup:**
    ```python
    # ... other configurations
    extensions = [
        "sphinx_js",
        # ... other extensions
    ]
    js_source_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "src"))
    js_language = 'typescript'
    jsdoc_config_path = "../typedoc.json"
    # Optional, but tried explicitly to confirm:
    js_node_executable = r"C:\Program Files\nodejs\node.exe"
    ```
3.  **Command:** Run `make clean && make html` from the `docs` directory.
4.  **Observe:** The build fails with the `WinError 193` exception.

### Expected Behavior

The documentation should build successfully for TypeScript source files, similar to how it builds for JavaScript. The Node.js subprocess should be invoked correctly by `sphinx-js`.

### Actual Behavior

The build process is terminated with an `OSError: [WinError 193]`, indicating that the subprocess call to Node.js was invalid.

### Additional Information

*   This error is consistent even after ensuring both Python and Node.js are 64-bit and hardcoding the `node.exe` path in `conf.py`.
*   Running `where node` on the affected system showed both `C:\Program Files\nodejs\node.exe` and `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\node.cmd`. Disabling the WindowsApps alias did not resolve the issue, suggesting a deeper conflict in the subprocess invocation.
*   The same setup works without issue on a Linux system, confirming the problem is specific to the Windows implementation of `sphinx-js`'s TypeDoc subprocess handling.
*   Disabling `js_language = 'typescript'` and relying on the JavaScript build works, indicating the core issue is limited to the TypeScript-specific codepath.

voxelstack:
Closes #285

I considered updating `search_node_modules` to return a command that may or may not include `"node"` but since including it fixes the Windows issue without breaking other platforms I went with this smaller change instead. This is also consistent with the `jsdoc`invocation where `"node"` is always part of the command.

On my windows machine the version call succeeds now but the project I'm testing it on is throwing an unrelated error about ESM URL schemes. Because of that I still haven't gotten a successful run after my changes, working on it now.

nox is also causing errors when I run it locally so after I get a successful run I'll try and check the tests locally with WSL.

### DIFF
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -27,8 +27,11 @@ MIN_TYPEDOC_VERSION = (0, 25, 0)
 
 @cache
 def typedoc_version_info(typedoc: str) -> tuple[tuple[int, ...], tuple[int, ...]]:
+    command = Command("node")
+    command.add(typedoc)
+    command.add("--version")
     result = subprocess.run(
-        [typedoc, "--version"],
+        command.make(),
         capture_output=True,
         encoding="utf8",
         check=True,
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -2,6 +2,7 @@
 
 import os
 import pathlib
+import posixpath
 import re
 import subprocess
 from collections.abc import Iterable, Iterator, Sequence
@@ -73,7 +74,7 @@ def typedoc_output(
     command.add("tsx@4.15.8")
     dir = Path(__file__).parent.resolve() / "js"
     command.add("--tsconfig", str(dir / "tsconfig.json"))
-    command.add("--import", str(dir / "registerImportHook.mjs"))
+    command.add("--import", "file:///" + str(dir / "registerImportHook.mjs"))
     command.add(str(dir / "main.ts"))
     if ts_sphinx_js_config:
         command.add("--sphinxJsConfig", ts_sphinx_js_config)
@@ -93,7 +94,11 @@ def typedoc_output(
     command.add("--excludePrivate", "false")
 
     with NamedTemporaryFile(mode="w+b", delete=False) as temp:
-        command.add("--json", temp.name, *abs_source_paths)
+        source_paths = abs_source_paths
+        if os.name == "nt":
+            source_paths = map(lambda path: str(posixpath.join(*str(path).split(os.sep))), abs_source_paths)
+    
+        command.add("--json", temp.name, *source_paths)
         try:
             subprocess.run(command.make(), check=True, env=env)
         except OSError as exc:
--- a/sphinx_js/typedoc.py
+++ b/sphinx_js/typedoc.py
@@ -96,8 +96,11 @@ def typedoc_output(
     with NamedTemporaryFile(mode="w+b", delete=False) as temp:
         source_paths = abs_source_paths
         if os.name == "nt":
-            source_paths = map(lambda path: str(posixpath.join(*str(path).split(os.sep))), abs_source_paths)
-    
+            source_paths = map(
+                lambda path: str(posixpath.join(*str(path).split(os.sep))),
+                abs_source_paths,
+            )
+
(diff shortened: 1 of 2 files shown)


voxelstack:
I guess there are more issues with Windows apart from running typedoc. What I'm seeing now is:
```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```
~~I suspect that must be caused by `importHooks.mjs` but the `fallbackUrl` starts with `file:` and `TYPEDOC_NODE_MODULES` is being set to the correct path. Having a little trouble with this one.~~
It's caused by paths being given to `--import`, so that is easily solvable by adding `"file:///" +` to the imports when running on Windows.

~~Now what I'm investigating is why the resolver is getting two absolute paths joined together: `'file:///D:/dev/ceres/docs/file:/C:/Users/voxel/pipx/venvs/sphinx-autobuild/Lib/site-packages/sphinx_js/js/main.ts'`~~

sphinx_js/typedoc.py
```diff
@@ -70,7 +74,7 @@ def typedoc_output(
     command.add("tsx@4.15.8")
     dir = Path(__file__).parent.resolve() / "js"
     command.add("--tsconfig", str(dir / "tsconfig.json"))
-    command.add("--import", str(dir / "registerImportHook.mjs"))
+    command.add("--import", "file:///" + str(dir / "registerImportHook.mjs"))
```
voxelstack, line 76: I feel like there must be a cleaner way to add a protocol to a path.

sphinx_js/typedoc.py
```diff
@@ -90,7 +94,14 @@ def typedoc_output(
     command.add("--excludePrivate", "false")
 
     with NamedTemporaryFile(mode="w+b", delete=False) as temp:
-        command.add("--json", temp.name, *abs_source_paths)
+        source_paths = abs_source_paths
+        if os.name == "nt":
+            source_paths = map(
+                lambda path: str(posixpath.join(*str(path).split(os.sep))),
```
voxelstack, line 99: I hate this, my first attempt was `path.split(os.sep)` but that errors out because `WindowsPath` doesn't have a split apparently.
Would love suggestions on how to make this better, the only thing that's important here is that these source paths use forward slashes.

voxelstack:
It works!

I'd appreciate help cleaning up the changes and running all checks locally but these changes were enough to get it working on my Windows machine.

hoodmane:
Can you add a windows CI job?

hoodmane:
And it looks like CI is broken so I'll have to fix that first...

voxelstack:
With some pointers I can look into why CI is broken, I just didn't know where to start since the error is on a file that wasn't changed.
I'm not familiar with most of the tools used on the checks and I want to do more research on them but haven't had a lot of time for it lately.
I marked it as ready for review because it's working for my use case but I will continue to work on it.

